PR #10, opened 2023-08-04
Add possibility to delete documents

alexander-schranz:
Currently it is not possible to delete a document. This will add possibility to remove documents by there identifier.

### DIFF
--- a/src/Internal/Engine.php
+++ b/src/Internal/Engine.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Loupe\Loupe\Internal;
 
+use Doctrine\DBAL\ArrayParameterType;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Driver\AbstractSQLiteDriver;
 use Doctrine\DBAL\Driver\API\SQLite\UserDefinedFunctions;
@@ -75,6 +76,25 @@ class Engine
         $indexer->addDocuments($documents);
 
         return $this;
+    }
+
+    /**
+     * @param array<int|string> $ids
+     */
+    public function deleteDocuments(array $ids): void
+    {
+        $this->getConnection()
+            ->executeStatement(
+                sprintf('DELETE FROM %s WHERE user_id IN(:ids)', IndexInfo::TABLE_NAME_DOCUMENTS),
+                [
+                    'ids' => array_map(function(string $id) {
+                        return LoupeTypes::convertToString($id);
+                    }, $ids),
+                ],
+                [
+                    'ids' => ArrayParameterType::STRING,
+                ]
+            );
     }
 
     public function countDocuments(): int
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+/vendor
+/composer.lock
--- a/src/Loupe.php
+++ b/src/Loupe.php
@@ -32,6 +32,22 @@ final class Loupe
         return $this;
     }
 
+    /**
+     * @throws IndexException
+     */
+    public function deleteDocument(int|string $id): void
+    {
+        $this->deleteDocuments([$id]);
+    }
+
+    /**
+     * @param array<int|string> $ids
+     */
+    public function deleteDocuments(array $ids): void
+    {
+        $this->engine->deleteDocuments($ids);
+    }
+
     public function countDocuments(): int
     {
         return $this->engine->countDocuments();
--- a/tests/Functional/IndexTest.php
+++ b/tests/Functional/IndexTest.php
@@ -139,6 +139,41 @@ class IndexTest extends TestCase
         $this->assertSame($uta, $document);
     }
 
+    public function testDeleteDocument(): void
+    {
+        $configuration = Configuration::create()
+            ->withFilterableAttributes(['departments', 'gender'])
+            ->withSortableAttributes(['firstname'])
+        ;
+
+        $loupe = $this->createLoupe($configuration);
+
+        $sandra = $this->getSandraDocument();
+        $uta = [
+            'id' => 2,
+            'firstname' => 'Uta',
+            'lastname' => 'Koertig',
+            'gender' => 'female',
+            'departments' => ['Development', 'Backoffice'],
+            'colors' => ['Red', 'Orange'],
+            'age' => 29,
+        ];
+
+        $loupe->addDocument($sandra);
+        $loupe->addDocument($uta);
+        $sandraDocument = $loupe->getDocument(1);
+        $utaDocument = $loupe->getDocument(2);
+        $this->assertSame($sandra, $sandraDocument);
+        $this->assertSame($uta, $utaDocument);
+
+        $loupe->deleteDocument(1);
+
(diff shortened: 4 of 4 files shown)


src/Internal/Engine.php
```diff
@@ -77,6 +78,25 @@ public function addDocuments(array $documents): self
...
+     * @param array<int|string> $ids
+     */
+    public function deleteDocuments(array $ids): void
+    {
+        $this->getConnection()
+            ->executeStatement(
+                sprintf('DELETE FROM %s WHERE user_id IN(:ids)', IndexInfo::TABLE_NAME_DOCUMENTS),
+                [
+                    'ids' => array_map(function(string $id) {
+                        return LoupeTypes::convertToString($id);
```
Toflar, line 91: You can use `convertToArrayOfStrings()` 😊

Toflar (changes requested):
Thanks for this addition! Unfortunately, we need to add quite a bit more logic here, not sure if you're up to it?

We not only need to cleanup the `documents` table but also 

* `multi_attributes_documents` for all documents that do not exist anymore
* `multi_attributes` for all attributes that do not exist in `multi_attributes_documents` anymore
* `terms_documents` for all documents that do not exist anymore
* `terms` for all terms that do not exist in `terms_documents` anymore
* After deleting, the IDF values need to be updated in order to have correct relevance sorting again.

Some of which is already in place in `updateInverseDocumentFrequencies()`. So imho, what needs to be done on top of your PR (let me know if you want to do it or not):

I would like to have 3 methods in the `Indexer`:

1. `private function reviseStorage()` - this should call `$this->removeOrphans();` and `$this->updateInverseDocumentFrequencies();`.
2. `private function removeOrphans()`
3. `private function updateInverseDocumentFrequencies()`

So what I would do is

1. Extract the one cleanup query in `updateInverseDocumentFrequencies()` into the new `removeOrphans` method.
2. Add the queries needed as mentioned above (the ones that should delete stuff from documents that do not exist anymore)
3. Call our new `$this->reviseStorage()` in `addDocuments()` and in your new `deleteDocuments()`.

That should get us our desired state 😊

Toflar:
Please, also target the `0.2` branch :)

alexander-schranz:
Could we in most cases not add foreign keys via activate `PRAGMA foreign_keys = ON;` on https://github.com/loupe-php/loupe/blob/405fee389ec3404f9ada4f72db2da9b4a7e42899/src/Internal/Engine.php#L57. Then I think we would only need todo:

 - `terms for all terms that do not exist in terms_documents anymore`
 
I'm not sure when I will come to it so if somebody have time before me finishing this feel free.

Toflar:
Not sure if `ON DELETE CASCADE` would solve this problem here. Most of the terms and attribute values will be used in multiple documents.

src/Internal/Index/Indexer.php
```diff
@@ -231,6 +255,43 @@ private function persistStateSet(): void
...
+    {
+        $this->removeOrphans();
+        $this->updateInverseDocumentFrequencies();
+    }
+
+    private function removeOrphans(): void
+    {
+        // Cleanup all terms documents which document not longer exists
+        $query = <<<'QUERY'
+            DELETE FROM %s WHERE document NOT IN (SELECT id FROM %s)
```
alexander-schranz, line 268: not sure about this as a FOREIGN KEY would be easier and the `id` selection would be very big here.

src/Internal/Index/Indexer.php
```diff
@@ -231,6 +255,43 @@ private function persistStateSet(): void
...
+            IndexInfo::TABLE_NAME_TERMS_DOCUMENTS,
+            IndexInfo::TABLE_NAME_DOCUMENTS,
+        );
+
+        $this->engine->getConnection()
+            ->executeQuery($query);
+
+        // Cleanup all multiple attributes documents which document not longer exists
+        $query = <<<'QUERY'
+            DELETE FROM %s WHERE document NOT IN (SELECT id FROM %s)
```
alexander-schranz, line 282: same as above

src/Internal/Index/Indexer.php
Toflar: Which would be optimized by SQLite, so you can really ignore this. But if you find a way with FKs, that's fine for me too.

alexander-schranz:
I'm not sure 100% if I did understand your comments correctly but I just pushed a version now. We probably should make some test case to make something like.

 - Add Document A
 - Add Document B
 - Calculate Table Counts
 - Add Document C
 - Delete Document C
 - Make sure all tables are as before add and deleting a document

src/Internal/Index/Indexer.php
```diff
@@ -231,6 +255,43 @@ private function persistStateSet(): void
...
+            IndexInfo::TABLE_NAME_DOCUMENTS,
+        );
+
+        $this->engine->getConnection()
+            ->executeQuery($query);
+    }
+
     private function updateInverseDocumentFrequencies(): void
     {
         // Cleanup all terms that are not in terms_documents anymore (to prevent division by 0)
```
Toflar, line 297: This cleanup, I would move to the `removeOprhans()` as well.

src/Internal/Index/Indexer.php
```diff
@@ -64,6 +65,29 @@ public function addDocuments(array $documents): self
...
+                    'ids' => array_map(function(string $id) {
+                        return LoupeTypes::convertToString($id);
+                    }, $ids),
+                ],
+                [
+                    'ids' => ArrayParameterType::STRING,
+                ]
+            );
+
+        $this->reviseStorage();
```
Toflar, line 86: This call also needs to go to the `addDocuments()` so it is the same :)

Toflar (commented):
I don't think we need to assert the table counts. I wanted to keep the connection internal so nobody tries to access it from the outside.
But essentially, what would be a problem, if we do not cleanup:

* Add document A
* Add document B
* Add document C
* Search ordered by `_relevance` (uses TF-IDF)
* Remove document A, B or C
* Search ordered by `_relevance` (uses TF-IDF) - this should now - depending on what terms were indexed and searched for, result in a different result as TF-IDF has changed. If, however, you just delete the documents and do not cleanup and re-calculate the document frequencies, this would not work properly.

So what we'd need to do is find an example for the term frequency ordering.

Toflar:
@alexander-schranz can you check if https://github.com/loupe-php/loupe/pull/27 fixes your problem?

alexander-schranz:
Think you wanted to comment on #12 :) but did test it and add some comments and edge cases via a test.

src/Internal/Engine.php
alexander-schranz: @Toflar did use `convertToArrayOfStrings` now, for that I needed to make the method public.

Toflar: Ah indeed, but I think that makes sense 👍

Toflar (approved):

Toflar:
Found a way to test the internal storage is updated using yet another new feature: https://github.com/loupe-php/loupe/pull/33

Finished your PR here - thanks Alex!

PR #11, opened 2023-08-04
Fix PHPStan level 0 problems

szepeviktor:
... and the text of a message.
Please see the diff.

@phpstan shows further problems.
```bash
vendor/bin/phpstan analyze src/ -l 4
```

### DIFF
--- a/src/Internal/CosineSimilarity.php
+++ b/src/Internal/CosineSimilarity.php
@@ -42,8 +42,8 @@ class CosineSimilarity
         $d2Norm = 0;
 
         foreach ($d1 as $i => $xi) {
-            if (isset($v2[$i])) {
-                $prod += $xi * $v2[$i];
+            if (isset($d2[$i])) {
+                $prod += $xi * $d2[$i];
             }
             $d1Norm += $xi * $xi;
         }
--- a/src/Internal/Filter/Parser.php
+++ b/src/Internal/Filter/Parser.php
@@ -127,7 +127,7 @@ class Parser
         $type = $token->type ?? null;
 
         if ($type === null || $type < 10 || $type > 30) {
-            $this->syntaxError('invalid operator', $token);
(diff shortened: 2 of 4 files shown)


src/Internal/CosineSimilarity.php
```diff
@@ -42,8 +42,8 @@ public static function similarity(array $d1, array $d2): float
         $d2Norm = 0;
 
         foreach ($d1 as $i => $xi) {
-            if (isset($v2[$i])) {
-                $prod += $xi * $v2[$i];
+            if (isset($d2[$i])) {
+                $prod += $xi * $d2[$i];
```
szepeviktor, line 46: A few line above I read

>      * Where d1 = query document (query terms)
>      * Where d2 = resulting document (matching terms)

You could name `d1` `queryDocument`, and `d2` `resultingDocument` and so on.

src/Internal/Filter/Parser.php
```diff
@@ -127,7 +127,7 @@ private function assertOperator(?Token $token): void
         $type = $token->type ?? null;
 
         if ($type === null || $type < 10 || $type > 30) {
-            $this->syntaxError('valid operator', $token);
+            $this->syntaxError('invalid operator', $token);
```
szepeviktor, line 130: I see :) the first parameter is called `$expected` - so it is correct.

Toflar:
Hi, thanks for the contribution. So do you actually want to do the work to add phpstan or not? Or what exactly is the purpose of this PR? :)

szepeviktor:
Glad to contribute.
This PR is an inspiration for you. But it includes fixing of two critical errors.

Toflar:
Okay, I know PHPStan but I'm closing this PR then as you don't plan to actually do the work. I don't see the point in having it float around here.

szepeviktor:
I am sorry to open it.

PR #12, opened 2023-08-04
Allow documents not contain all fields

alexander-schranz:
It can be that documents maybe don't contain all fields like here:

https://github.com/schranz-search/schranz-search/blob/3b0660babbddee6c47d297e01e0fdde425e49f56/packages/seal/src/Testing/TestingHelper.php#L178-L193

### DIFF
--- a/src/Internal/Index/IndexInfo.php
+++ b/src/Internal/Index/IndexInfo.php
@@ -195,10 +195,7 @@ class IndexInfo
     public function getLoupeTypeForAttribute(string $attributeName): string
     {
         if (! \array_key_exists($attributeName, $this->getDocumentSchema())) {
-            throw new InvalidConfigurationException(sprintf(
-                'The attribute "%s" does not exist on the document schema.',
-                $attributeName
-            ));
+            return LoupeTypes::TYPE_NULL;
         }
 
         return $this->getDocumentSchema()[$attributeName];
--- a/tests/Functional/IndexTest.php
+++ b/tests/Functional/IndexTest.php
@@ -7,6 +7,7 @@ namespace Loupe\Loupe\Tests\Functional;
 use Loupe\Loupe\Configuration;
 use Loupe\Loupe\Exception\LoupeExceptionInterface;
 use Loupe\Loupe\Logger\InMemoryLogger;
+use Loupe\Loupe\SearchParameters;
 use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Filesystem\Filesystem;
@@ -168,6 +169,11 @@ class IndexTest extends TestCase
         $loupe = $this->createLoupe($configuration);
         $loupe->addDocuments($documents);
         $this->assertSame(\count($documents), $loupe->countDocuments());
+
+        $searchParameters = SearchParameters::create()
+            ->withFilter('departments = \'Development\'');
+
+        $this->assertGreaterThanOrEqual(1, $loupe->search($searchParameters)->getHits());
     }
 
     public static function validSchemaChangesProvider(): \Generator
--- a/tests/Functional/IndexTest.php
+++ b/tests/Functional/IndexTest.php
@@ -223,6 +223,13 @@ class IndexTest extends TestCase
                 ]),
             ],
         ];
+
+        yield 'Omitting filterable attributes in first document allowed' => [
+            [
+                \array_filter(\array_merge(self::getUtaDocument(), ['departments' => null, 'gender' => null])),
+                self::getSandraDocument(),
+            ],
+        ];
     }
 
     /**


phpunit.xml.dist
```diff
@@ -4,6 +4,8 @@
          backupGlobals="false"
          bootstrap="vendor/autoload.php"
          colors="true"
+         displayDetailsOnTestsThatTriggerWarnings="true"
+         failOnWarning="true"
```
alexander-schranz, line 8: Without following errors were not shown:

Bildschirmfoto 2023-08-05 um 00 30 48

src/Internal/Index/IndexInfo.php
```diff
@@ -207,14 +207,6 @@ public function validateDocument(array $document): void
     {
         $documentSchema = $this->getDocumentSchema();
 
-        if (\count(array_diff_key($documentSchema, $document)) !== 0) {
-            throw InvalidDocumentException::becauseDoesNotMatchSchema(
-                $documentSchema,
-                $document,
-                $document[$this->engine->getConfiguration()->getPrimaryKey()] ?? null
-            );
-        }
```
alexander-schranz, line 216: As I think all fields could be empty we don't need this check. Another possibility would be a prefill this way:

```php
        $missingColumns = array_diff_key($documentSchema, $document);

        if (count($missingColumns) !== 0) {
            foreach ($missingColumns as $missingColumn => $type) {
                $document[$missingColumn] = LoupeTypes::isSingleType($type) ? null : [];
            }
        }
```

But as I think its common that maybe a field is later added I think we can just ignore the missing fields.

src/Internal/Index/Indexer.php
```diff
@@ -116,6 +116,8 @@ private function indexDocument(array $document): int
                 continue;
             }
 
+            $attributeValue = $document[$attribute] ?? null;
```
szepeviktor, line 119: The base value of an array is an empty array.
Please consider using `?? []` in the method too.

[_source_](https://github.com/szepeviktor/very-clean-code#-base-value-of-an-array)

szepeviktor: ... loupe may expect that its base value is `null`

alexander-schranz: this is looped through all attributes which mean it can be a single value (null)  or a multi value (null). Currently don't see any problem with null atleast my and the internal loupe test didn't fail. Do you have a test case which we maybe need to add where you think this need a more complex logic to check what the better default value is based on the type?

szepeviktor: > Do you have a test case

No. I just encourage people to use non-nullable arrays.

alexander-schranz: in this case the variables can be various values not only a array e.g. `firstName => null` makes more sense than `firstName => []`.

szepeviktor: Of course!
Only arrays' default value should be `[]`

phpunit.xml.dist
Toflar: I've added this to the 0.2 branch.

src/Internal/Index/Indexer.php
Toflar: `null` is the correct value. Currently, Loupe does not make a difference between `[]` and `null` yet so it doesn't matter. But once we introduce filters for `IS EMPTY` and `IS NULL`, we will need this difference. So if it hasn't been provided, it should definitely be `null`.

src/Internal/Index/IndexInfo.php
Toflar: If we would prefill them with `null` here, we could get rid of all the other checks you added. So we'd have it in one central place rather than scattered all over the code base, right?

Toflar (commented):
Certainly looks good to me - can you also target the 0.2 branch, please?

alexander-schranz:
> If we would prefill them with null here, we could get rid of all the other checks you added. So we'd have it in one central place rather than scattered all over the code base, right?

I think not in all cases. As if later new fields are added we still need that `?? null` checks at they could be indexed before without all fields.

Toflar:
No, you cannot add new attributes to a document which was previously not present. That's not a use case I want to allow.

Toflar:
In other words, all attributes always have to be present. Actually I don't want to allow only providing a subset - the developers should ensure they pass `null` values themselves.

alexander-schranz:
@Toflar okay then it would work to prefill it at: https://github.com/loupe-php/loupe/pull/12#discussion_r1284891188. Or would you prefer close this pull request then?

Then I need to handle it in SEAL. I know I had it once in SEAL but typesense didn't like null values then I removed that 😄

Toflar:
> Then I need to handle it in SEAL. I know I had it once in SEAL but typesense didn't like null values then I removed that 😄

My question for other search engines would be, whether this is possible:

```php
$engine->addDocument(['id' => 42, 'attribute_1' => 'foobar', 'attribute_2' => 'foobar 2']); // first call, defines the schema
$engine->addDocument(['id' => 42, 'attribute_1' => 'foobar']);
```

If yes, I think we should also allow it. What I actually don't really want to allow is this:

```php
$engine->addDocument(['id' => 42, 'attribute_1' => 'foobar']); // first call, defines the schema
$engine->addDocument(['id' => 42, 'attribute_1' => 'foobar', 'attribute_2' => 'foobar 2']);
```

alexander-schranz:
@Toflar 

> $engine->addDocument(['id' => 42, 'attribute_1' => 'foobar']); // first call, defines the schema
> $engine->addDocument(['id' => 42, 'attribute_1' => 'foobar', 'attribute_2' => 'foobar 2']);

New fields mostly are not the problem as I know. Keep in mind in `SEAL` we always have a fixed schema and not a `dynamic` one. As example in Elasticsearch already the Schema is created in the SchemaManager::createIndex with all fields. Still if we have a look at dynamic Elasticsearch default, adding new fields will just be guesed and be indexed. So something like following works in Elasticsearch:

```
$engine->addDocument(['id' => 42, 'attribute_1' => 'foobar']); // first call, defines the schema
$engine->addDocument(['id' => 42, 'attribute_1' => 'foobar', 'attribute_2' => 'foobar 2']);
```

What errors is changing the type and errors as example when the dynamic mapping is first indexes as numeric based on this mapping https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-field-mapping.html but then changed to another type:

```
$engine->addDocument(['id' => 42, 'attribute_1' => 'foobar']); // first call, defines the schema
$engine->addDocument(['id' => 42, 'attribute_1' => 'foobar', 'attribute_2' => 1.5]);
$engine->addDocument(['id' => 42, 'attribute_1' => 'foobar', 'attribute_2' => 'foobar 2']); // errors attribute_2 field was created as number
```

alexander-schranz:
@Toflar here is also the link with the different strategies elasticsearch / opensearch support for new added fields: https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic.html#dynamic-parameters. As written above in SEAL we forcing `strict` schema currently. For the other engines I'm not aware if a dynamic field is searchable or not.

Toflar:
I have reworked schema handling in https://github.com/loupe-php/loupe/pull/24. Can you check if the documentation makes sense and fixes the issues you're having?

alexander-schranz:
@Toflar 

> ## Omitting an attribute

Works 👍 

> ## Adding a new attribute

Fails. See `IndexTest::testSchemaAddNewAttributes`

Toflar:
@alexander-schranz I think we can close this now as all the bugs have been fixed in the 0.2 branch?

PR #14, opened 2023-08-05
Add support for NOT Equals != operator

alexander-schranz:
Add not equals operator.

And fix NOT IN operator.

--> own pull request fixed in #20 

Currently the negative operators do not work as expected, when try to filter out all which matches a specific value they are still shown aslong as they have another value which should not be the case.

Example in SEAL:

| Id | Tags        | Expected | Current     |  Why?
|----|-------------|----------|-------------|-----------
| 1  | Tech<br> UI |          |   X         | Has one none -> logic need to be inverted
| 2  | UI<br>UX    |          |   X         | Has one none -> logic need to be inverted
| 3  | Tech<br>UX  |    X    |   X          | 
| 4  |             |    X    |              | Empty currently filter out -> logic need to be inverted

The query is `tags != 'UI'` and the expected result is `3` and `4`, but the result is: `1`, `2`, `3`. Same currently appears for the `NOT IN` filter I adopted the existing test cases there as they still are returning which should filtered out.

Also adopted the test cases for the expected result.

## Solution

The current SQL looks like this:

```sql
// != example
SELECT COUNT() OVER() AS totalHits, d.document
FROM documents d WHERE d.id IN (
    SELECT document FROM multi_attributes_documents mad
    INNER JOIN multi_attributes ma
        ON ma.attribute=:dcValue1
        AND ma.id = mad.attribute
    WHERE ma.string_value != 'Backoffice'
)
ORDER BY d.firstname ASC LIMIT 20

// NOT IN example
SELECT COUNT() OVER() AS totalHits, d.document
FROM documents d WHERE d.id IN (
    SELECT document FROM multi_attributes_documents mad
    INNER JOIN multi_attributes ma
        ON ma.attribute=:dcValue1
        AND ma.id = mad.attribute
    WHERE ma.string_value NOT IN ('Backoffice', 'Project Managemen')
)
ORDER BY d.firstname ASC LIMIT 20
```

But to get the expected result we would require to change for `NOT` statements (NOT IN and !=) the current SQL which looks like this:

```sql
// != example
SELECT COUNT() OVER() AS totalHits, d.document
FROM documents d WHERE d.id NOT IN (  // <--- replace IN with NOT IN
    SELECT document FROM multi_attributes_documents mad
    INNER JOIN multi_attributes ma
        ON ma.attribute=:dcValue1
        AND ma.id = mad.attribute
    WHERE ma.string_value = 'Backoffice' // <--- replace != with =
)
ORDER BY d.firstname ASC LIMIT 20

// NOT IN example
SELECT COUNT() OVER() AS totalHits, d.document
FROM documents d WHERE d.id NOT IN (  // <--- replace IN with NOT IN
    SELECT document FROM multi_attributes_documents mad
    INNER JOIN multi_attributes ma
        ON ma.attribute=:dcValue1
        AND ma.id = mad.attribute
    WHERE ma.string_valu IN ('Backoffice', 'Project Managemen') // <--- replace != with =
)
ORDER BY d.firstname ASC LIMIT 20
```

## TODO

 - [ ] We should maybe add test fixtures somebody which have maybe no `departments` currently empty multi value is not tested.

### DIFF
--- a/tests/Functional/SearchTest.php
+++ b/tests/Functional/SearchTest.php
@@ -15,6 +15,81 @@ class SearchTest extends TestCase
 {
     use FunctionalTestTrait;
 
+    public static function equalFilterProvider(): \Generator
+    {
+        yield '= on multiple attribute' => [
+            "departments = 'Backoffice'",
+            [
+                [
+                    'id' => 6,
+                    'firstname' => 'Huckleberry',
+                ],
+                [
+                    'id' => 5,
+                    'firstname' => 'Marko',
+                ],
+                [
+                    'id' => 2,
+                    'firstname' => 'Uta',
+                ],
+            ],
+        ];
+
+        yield '!= on multiple attribute' => [
+            "departments != 'Backoffice'",
+            [
+                [
+                    'id' => 3,
+                    'firstname' => 'Alexander',
+                ],
+                [
+                    'id' => 4,
+                    'firstname' => 'Jonas',
+                ],
+                [
+                    'id' => 1,
+                    'firstname' => 'Sandra',
+                ],
+            ],
+        ];
+
+        yield '= on single attribute' => [
+            "gender = 'female'",
+            [
+                [
+                    'id' => 1,
+                    'firstname' => 'Sandra',
+                ],
+                [
+                    'id' => 2,
+                    'firstname' => 'Uta',
+                ],
+            ],
+        ];
+
+        yield '!= on single attribute' => [
+            "gender != 'female'",
+            [
+                [
+                    'id' => 3,
+                    'firstname' => 'Alexander',
+                ],
+                [
+                    'id' => 6,
+                    'firstname' => 'Huckleberry',
+                ],
+                [
+                    'id' => 4,
+                    'firstname' => 'Jonas',
+                ],
+                [
+                    'id' => 5,
+                    'firstname' => 'Marko',
+                ],
+            ],
+        ];
+    }
+
     public static function highlightingProvider(): \Generator
     {
         yield 'Highlight with matches position only' => [
@@ -270,6 +345,27 @@ class SearchTest extends TestCase
             'page' => 1,
             'totalPages' => 1,
             'totalHits' => 1,
+        ]);
+    }
+
+    #[DataProvider('equalFilterProvider')]
+    public function testEqualFilter(string $filter, array $expectedHits): void
+    {
+        $loupe = $this->setupLoupeWithDepartmentsFixture();
+
+        $searchParameters = SearchParameters::create()
+            ->withAttributesToRetrieve(['id', 'firstname'])
+            ->withFilter($filter)
+            ->withSort(['firstname:asc'])
+        ;
+
+        $this->searchAndAssertResults($loupe, $searchParameters, [
+            'hits' => $expectedHits,
+            'query' => '',
+            'hitsPerPage' => 20,
+            'page' => 1,
+            'totalPages' => 1,
+            'totalHits' => \count($expectedHits),
         ]);
     }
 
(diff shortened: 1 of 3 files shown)


tests/Functional/SearchTest.php
```diff
@@ -156,18 +231,10 @@ public static function inFilterProvider(): \Generator
                     'id' => 3,
                     'firstname' => 'Alexander',
                 ],
-                [
-                    'id' => 4,
-                    'firstname' => 'Jonas',
```
alexander-schranz, line 161: `Jonas` is part of `Project Management` and should not be shown.

tests/Functional/SearchTest.php
```diff
@@ -156,18 +231,10 @@ public static function inFilterProvider(): \Generator
...
-                ],
                 [
                     'id' => 1,
                     'firstname' => 'Sandra',
                 ],
-                [
-                    'id' => 2,
-                    'firstname' => 'Uta',
```
alexander-schranz, line 169: `Uta` is part of `Backoffice` and should not be shown

src/Internal/Search/Searcher.php
```diff
@@ -404,9 +404,14 @@ private function handleFilterAstNode(Node $node, array &$whereStatement): void
...
             if (\in_array($node->attribute, $this->engine->getIndexInfo()->getMultiFilterableAttributes(), true)) {
-                $whereStatement[] = $documentAlias . '.id IN (';
+                // @see https://github.com/loupe-php/loupe/pull/14
+                $previousOperator = $node->operator;
+                $operator = $node->operator->getMultiValueOperator();
+                $node->operator = $operator;
+                $whereStatement[] = $documentAlias . '.id ' . ($node->operator !== $previousOperator ? 'NOT IN' : 'IN') . ' (';
                 $whereStatement[] = $this->createSubQueryForMultiAttribute($node);
                 $whereStatement[] = ')';
+                $node->operator = $previousOperator;
```
alexander-schranz, line 414: I'm not happy with this code but maybe you have a better idea.

tests/Functional/SearchTest.php
```diff
@@ -15,6 +15,81 @@ class SearchTest extends TestCase
 {
     use FunctionalTestTrait;
 
+    public static function equalFilterProvider(): \Generator
+    {
+        yield '= on multiple attribute' => [
```
Toflar, line 20: Do you know if applying a single value operator like `=` or `!=` works on array-value attributes in other search engines? Does this work in MeiliSearch for example?

Toflar (commented):
I think we should split fixing the `NOT IN` filter and adding support for `!=` (and applying those on multi attributes (to be checked if that's normal first)) in 2 PRs because they are not really related.

tests/Functional/SearchTest.php
alexander-schranz: @Toflar yes `=` and `!=` on multi values did work in all other engines yet which are supported by SEAL.

Toflar: Okay, then let's add support for this in a separate PR. We should first fix the `NOT IN` logic and then add support for single comparisons on multi attributes. We should then also check what does happen when I apply `>=` on multi attributes.

alexander-schranz: I can tackle the split at the evening. I could check what the case for `>=` and `<=` is in other engines on multi values and give you an update.

alexander-schranz: @Toflar Fix for `NOT IN`: https://github.com/loupe-php/loupe/pull/19

alexander-schranz: @Toflar `>=` and `<=` works on multi values also added additional test cases and the did work on currently all supported engines. https://github.com/schranz-search/schranz-search/pull/226

Toflar:
Thanks, Alex!

PR #20, opened 2023-08-08
Fixed incorrect negative search on multi attribute values

Toflar:
Alternative to https://github.com/loupe-php/loupe/pull/19

### DIFF
--- a/src/Internal/Filter/Ast/Operator.php
+++ b/src/Internal/Filter/Ast/Operator.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Loupe\Loupe\Internal\Filter\Ast;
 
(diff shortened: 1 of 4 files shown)


src/Internal/Filter/Ast/Operator.php
```diff
@@ -14,9 +15,10 @@ enum Operator: string
     case In = 'IN';
     case LowerThan = '<';
     case LowerThanOrEquals = '<=';
+    case NotEquals = '!=';
```
alexander-schranz, line 18: should this not be in an own pull request 😄

Toflar: Yeah I wanted to have all the opposites avaiable in this PR. It still needs to be supported in the parser though.

PR #24, opened 2023-08-09
Reworked and documented schema handling

Toflar:


alexander-schranz:
Text does make sense for me 👍 For testing against seal I need first rebase my branch, but can give you maybe an update later.

PR #27, opened 2023-08-14
Fixed filtering and sorting on not-yet existing schema values

Toflar:

PR #28, opened 2023-08-14
Add TestCase for added filterable attribute

alexander-schranz:
This test some edge cases of adding and removing fields and adopt the Indexer hackly to support that cases.


Toflar:
This should be covered in https://github.com/loupe-php/loupe/pull/27 now. Thanks for the test case - don't worry, I'll co-author you in https://github.com/loupe-php/loupe/pull/27 :)

PR #29, opened 2023-08-14
Add missing test cases for greater and lower than filters

alexander-schranz:
Greater Than and Less Than for a attribute which is maybe not set or null unexpectly different to other search engines and natural expected behaviour.


Toflar:
I've merged your commit into https://github.com/loupe-php/loupe/pull/31 and fixed the problem there :)

PR #30, opened 2023-08-15
Add test case for sort by asc and des with null and missing values

alexander-schranz:
Sorting for a attribute which is maybe not set or null unexpectly different to other search engines and natural behaviour.

Ported this test case from SEAL to Loupe here.


Toflar:
Fixed by https://github.com/loupe-php/loupe/pull/32.

PR #44, opened 2023-08-26
Set min requirement for doctrine dbal

alexander-schranz:
3.5.0 does not have the ArrayParameter class this does only exist since 3.6.

Bildschirmfoto 2023-08-26 um 17 23 28


alexander-schranz:
@Toflar think such errors we could avoid to run one CI task on prefer-lowest?

Toflar:
On it already in https://github.com/loupe-php/loupe/pull/45.

PR #50, opened 2023-09-06
Logo

doishub:
Adds the logo/emblem as an asset and uses it in the readme (#2).


Toflar:
Can you add `/logo` to `.gitattributes` so we don't ship it via Composer, please? 😊

Toflar:
Thank you so much - I love it! ❤️

Issue #55, opened 2024-01-17
Highlighter doesn't add end tag


MaximilianKresse:
Hi. I discovered a strange behaviour regarding the highlight feature of the search. 
In some constellations the end tag of "em" isn't added to the formatted string.

```php
<?php

use Loupe\Loupe\Configuration;
use Loupe\Loupe\Loupe;
use Loupe\Loupe\LoupeFactory;
use Loupe\Loupe\SearchParameters;

require_once(__DIR__ . '/vendor/autoload.php');

$configuration = Configuration::create()
    ->withPrimaryKey('id')
    ->withSearchableAttributes(['Title'])
;

$loupeFactory = new LoupeFactory();
$loupe = $loupeFactory->create(__DIR__ . '/loupe', $configuration);

$loupe->addDocument([
    'id' => 1,
    'Title' => 'Hallo Welt',
]);

$searchParameters = SearchParameters::create()
    ->withQuery('Welt')
    ->withAttributesToSearchOn(['Title'])
    ->withAttributesToHighlight(['Title'])
;
$results = $loupe->search($searchParameters);
$hit = $results->getHits()[0];

var_dump($hit['_formatted']['Title']);
```

Expected result:
```
string(19) "Hallo <em>Welt</em>"
```

Actual result:
```
string(14) "Hallo <em>Welt"
```

The problem seem to occur if the matched result is at the end of the field.

Toflar:
Closing in favor of #56 :)

PR #56, opened 2024-01-17
Fixed highlighting #55

MaximilianKresse:
See #55

### DIFF
--- a/src/Internal/Search/Highlighter/Highlighter.php
+++ b/src/Internal/Search/Highlighter/Highlighter.php
@@ -60,6 +60,10 @@ class Highlighter
             $highlightedText .= $char;
         }
 
+        if (\in_array(UTF8::strlen($text), $spans['ends'])) {
+            $highlightedText .= $endTag;
+        }
+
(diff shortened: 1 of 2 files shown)


src/Internal/Search/Highlighter/Highlighter.php
```diff
@@ -60,6 +60,10 @@ public function highlight(string $text, TokenCollection $queryTokens): Highlight
             $highlightedText .= $char;
         }
 
+        if (\in_array(UTF8::strlen($text), $spans['ends'])) {
```
Toflar, line 63:
```suggestion
        // Match at the end of the $text
        if (\in_array($pos + 1, $spans['ends'])) {
```

Toflar (changes requested):
Hey, nice find - thanks for the PR! ❤️ 
I think we could just re-use `$pos` as we already split the string, we don't need to waste performance on re-counting, right?

Toflar:
Released as 0.5.4 - thanks!

PR #70, opened 2024-01-30
Add possibility to define highlight start and end tag

alexander-schranz:
Currently always `<em></em>` is used for the highlighted text but `<mark></mark>` or custom tags with classes are maybe in some cases used.

Currently looking into how to implement such things into SEAL: https://github.com/schranz-search/schranz-search/issues/343

### DIFF
--- a/src/Internal/Search/Highlighter/Highlighter.php
+++ b/src/Internal/Search/Highlighter/Highlighter.php
@@ -17,8 +17,12 @@ class Highlighter
     ) {
     }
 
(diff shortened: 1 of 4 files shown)


Toflar (changes requested):
Nice one! ❤️  This is a new feature and should be rebased onto `develop`. There will be merge conflicts because there are already other improvements on highlighting for the next version 😊

Toflar (approved):

Toflar:
Thanks @alexander-schranz !

alexander-schranz:
Nice, thx for the quick review :)

PR #82, opened 2024-09-09
Add more test cases for distance checking

alexander-schranz:
Build ontop of #81 #70

### DIFF
--- a/tests/Functional/SearchTest.php
+++ b/tests/Functional/SearchTest.php
@@ -1166,7 +1166,7 @@ class SearchTest extends TestCase
                         'lat' => 51.5074,
                         'lng' => -0.1278,
                     ],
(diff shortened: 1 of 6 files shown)


tests/Functional/SearchTest.php
```diff
@@ -26,6 +26,15 @@ public static function distanceFilterProvider(): iterable
         yield [
             'distance' => 4_587_759,
         ];
+        yield [
+            'distance' => 4_642_695,
+        ];
+        yield [
+            'distance' => 4_642_696, // TODO all distanced over >= 4_642_696 does not work
```
alexander-schranz, line 33: if we can not support this long distances other possibility would maybe be `outofboundedexception` or something like that 🤔

Toflar:
All fixed and released.

PR #84, opened 2024-09-17
Add GeoBoundingBox filter

### LINKED ISSUE #83: Support for filtering by a GeoBoundingBox
Why Loupe already supports filtering by a GeoDistance. Which support we already merged also int SEAL. It would also be nice to filter by a bounding box (rectangle). That is mostly used for things when a user navigates around a map and have a specific area shown.

Example for a bounding box search AirBnB:

Bildschirmfoto 2024-09-16 um 09 36 00

---

I did some research about it what input parameters most search engine uses:

**Algolia**, **Elasticsearch** / **Opensearch** are using Top Left -> Bottom Right ↘️ 

**Meilisearch** is using Top Right -> Bottom Left ↙️

For me it was little bit strange and unusually that Meilisearch did go here another way. But after some research I found out it totally make sense. The JS libraries [Google Maps API](https://developers.google.com/maps/documentation/javascript/reference/coordinates#LatLngBounds) and [MapBox API](https://docs.mapbox.com/help/glossary/bounding-box/) returns NorthEast (TopRight) -> SouthWest (BottomLeft). https://github.com/meilisearch/specifications/pull/223. I like that way to be similar to the JS libraries most people will use and think that make more sense to avoid converting to another way.

So I think the `$north, $east, $south, $west` or `[$north, $east], [$south, $west]` make most sense for input parameters.

Links to the different search engines docs about this can be found here: https://github.com/schranz-search/schranz-search/issues/365#issuecomment-2351592129.

---

Some search engine only support polygon. That might be a problem if you view on areas which are bigger as half the world map as polygon goes then the shorter way and than completly exclude the area you view. In a polygon based search on areas bigger half worldmap we would need fill things with additional coordinates to avoid that flip of the polygon.

alexander-schranz:
Why Loupe already supports filtering by a GeoDistance. Which support we already merged also int SEAL. It would also be nice to filter by a bounding box (rectangle). That is mostly used for things when a user navigates around a map and have a specific area shown.

See fore more information: #83 

fixes #83

Implemented as far as I could get with my knowledge.

### DIFF
--- a/docs/searching.md
+++ b/docs/searching.md
@@ -286,11 +286,11 @@ $results = [
 
 Additional to a query based on distance we can also search for locations inside a bounding box.
 In this example we have 4 documents and 3 with geo coordinates (New York, London, Vienna).
-We have a bounding context which goes from Dublin to Athen which so returns only London and Vienna.
+We create a bounding box filter which spans from Dublin to Athens which then only matches our documents in London and Vienna.
 
 Keep in mind that the order of the arguments is important.
-The `_geoBoundingBox` expect `attributeName`, `north` (top), `east` (right), `south` (bottom), `west` (left).
-In this specific example top is Dublin Latitude, right is Athen Longitude, bottom is Athen Latitude and left is Dublin Longitude.
+The `_geoBoundingBox` expects `attributeName`, `north` (top), `east` (right), `south` (bottom), `west` (left).
+In this specific example, `top` is the latitude of Dublin,`right` is the longitude of Athens,`bottom` is the latitude of Athens and`left` equals Dublin's longitude.
 
 ```php
 $searchParameters = SearchParameters::create()
@@ -299,7 +299,7 @@ $searchParameters = SearchParameters::create()
 ;
 ```
 
-This will result looks like this::
+This is going to be your result:
 
 ```php
 $results = [
--- a/docs/searching.md
+++ b/docs/searching.md
@@ -283,3 +283,51 @@ $results = [
 
 [Config]: configuration.md
 [Restaurant_Fixture]: ./../tests/Functional/IndexData/restaurants.json
+
+Additional to a query based on distance we can also search for locations inside a bounding box.
+In this example we have 4 documents and 3 with geo coordinates (New York, London, Vienna).
+We have a bounding context which goes from Dublin to Athen which so returns only London and Vienna.
+
+Keep in mind that the order of the arguments is important.
+The `_geoBoundingBox` expect `attributeName`, `north` (top), `east` (right), `south` (bottom), `west` (left).
+In this specific example top is Dublin Latitude, right is Athen Longitude, bottom is Athen Latitude and left is Dublin Longitude.
+
+```php
+$searchParameters = SearchParameters::create()
+    ->withAttributesToRetrieve(['id', 'name', 'location'])
+    ->withFilter('_geoBoundingBox(location, 53.3498, 23.7275, 37.9838, -6.2603)')
+;
+```
+
+This will result looks like this::
+
+```php
+$results = [
+    'hits' => [
+        [
+            'id' => '2',
+            'title' => 'London',
+            'location' => [
+                'lat' => 51.5074,
+                'lng' => -0.1278,
+            ],
+        ],
+        [
+            'id' => '3',
+            'title' => 'Vienna',
+            'location' => [
+                'lat' => 48.2082,
+                'lng' => 16.3738,
+            ],
+        ],
+    ],
+    'query' => '',
+    'hitsPerPage' => 20,
+    'page' => 1,
+    'totalPages' => 1,
+    'totalHits' => 2,
+];
+```
+
+[Config]: configuration.md
+[Restaurant_Fixture]: ./../tests/Functional/IndexData/locations.json
--- a/src/Internal/Filter/Ast/GeoBoundingBox.php
+++ b/src/Internal/Filter/Ast/GeoBoundingBox.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Loupe\Loupe\Internal\Filter\Ast;
+
+use Location\Bounds;
+use Location\Coordinate;
+
+class GeoBoundingBox extends Node
+{
+    public function __construct(
+        public string $attributeName,
+        public float $north,
+        public float $east,
+        public float $south,
+        public float $west,
+    ) {
+    }
+
+    public function getBbox(): Bounds
+    {
+        // phpgeo bounds are top left to bottom right but meilisearch and so loupe is top right to bottom left
+        return new Bounds(
+            new Coordinate($this->north, $this->west),
+            new Coordinate($this->south, $this->east),
+        );
+    }
+
+    public function toArray(): array
+    {
+        return [
+            'attribute' => $this->attributeName,
+            'north' => $this->north,
+            'east' => $this->east,
+            'south' => $this->south,
+            'west' => $this->west,
+        ];
+    }
+}
--- a/src/Internal/Filter/Ast/GeoBoundingBox.php
+++ b/src/Internal/Filter/Ast/GeoBoundingBox.php
@@ -20,7 +20,6 @@ class GeoBoundingBox extends Node
 
     public function getBbox(): Bounds
     {
-        // phpgeo bounds are top left to bottom right but meilisearch and so loupe is top right to bottom left
         return new Bounds(
             new Coordinate($this->north, $this->west),
             new Coordinate($this->south, $this->east),
--- a/src/Internal/Filter/Ast/GeoBoundingBox.php
+++ b/src/Internal/Filter/Ast/GeoBoundingBox.php
@@ -9,31 +9,34 @@ use Location\Coordinate;
 
 class GeoBoundingBox extends Node
 {
+    private Bounds $bbox;
+
     public function __construct(
         public string $attributeName,
-        public float $north,
-        public float $east,
-        public float $south,
-        public float $west,
+        float $north,
+        float $east,
+        float $south,
+        float $west,
     ) {
+        $this->bbox = new Bounds(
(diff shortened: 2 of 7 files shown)


tests/Functional/SearchTest.php
```diff
@@ -1187,6 +1187,60 @@ public function testGeoSearchRetrieveDistanceWithoutSort(): void
...
+
+        $loupe = $this->createLoupe($configuration);
+        $this->indexFixture($loupe, 'locations');
+
+        $dublin = ['lat' => 53.3498, 'lng' => -6.2603];
+        $athen = ['lat' => 37.9838, 'lng' => 23.7275];
+
+        $searchParameters = SearchParameters::create()
+            ->withFilter(\sprintf(
+                '_geoBoundingBox(%s, [%s, %s], [%s, %s])',
```
alexander-schranz, line 1205: See https://www.meilisearch.com/docs/learn/filtering_and_sorting/geosearch#_geoboundingbox which uses `[]` syntax. Not sure if we may want to go with `_geoBoundingBox(%s, %s, %s, %s, %s)`

src/Internal/Filter/Ast/GeoBoundingBox.php
```diff
@@ -0,0 +1,41 @@
...
+use Location\Coordinate;
+
+class GeoBoundingBox extends Node
+{
+    public function __construct(
+        public string $attributeName,
+        public float $north,
+        public float $east,
+        public float $south,
+        public float $west,
```
alexander-schranz, line 18: another possibility would be name it top, right, left, bottom

alexander-schranz: In the meilisearch docs they only reference: `_geoBoundingBox([{lat}, {lng}], [{lat}, {lng}])`

But in the code its topRight, bottomLeft: https://github.com/meilisearch/meilisearch/blob/93f0317b9467c96e0332024450bcaf557c5bc9b4/filter-parser/src/lib.rs#L147

Toflar: Internally, we can work with the classes from `phpgeo`. I guess you can work with `Bounds` like we do in `GeoDinstance`, right?

tests/Functional/SearchTest.php
Toflar: I think it's fine to only support the easier variant for now :)

src/Internal/Filter/Parser.php
```diff
@@ -263,6 +270,23 @@ private function handleGeoRadius(Engine $engine): void
...
 
+    private function handleGeoBoundingBox(Engine $engine): void
+    {
+        $this->assertOpeningParenthesis($this->lexer->lookahead);
+        $this->lexer->moveNext();
+        $this->lexer->moveNext();
+
+        $attributeName = (string) $this->lexer->token?->value;
+
+        // TODO
```
Toflar, line 281: You are almost there! Look at `handleGeoRadius`. Basically you need a few `$this->assertAndExtractFloat($this->lexer->token, true);`, `$this->assertComma($this->lexer->lookahead);` and `$this->lexer->moveNext()` calls.
I always start with adding the tests to `ParserTest` and then adjust the `Parser` until everything is green ;)

Toflar (commented):
> Implemented as far as I could get with my knowledge.

Do you want me to take over from here or do you want to give it a shot with my feedback? You are almost there!

src/Internal/Filter/Ast/GeoBoundingBox.php
alexander-schranz: I return a Bound object below to work with that later

tests/Functional/SearchTest.php
alexander-schranz: 👍

alexander-schranz:
@Toflar ready for review

src/Internal/Search/Searcher.php
```diff
@@ -611,6 +612,51 @@ private function handleFilterAstNode(Node $node, array &$whereStatement): void
             $whereStatement[] = ')';
         }
 
+        if ($node instanceof GeoBoundingBox) {
```
Toflar, line 615: Isn't this duplicate code? Looks almost like the `GeoDistance` logic - can't we extract that in a `private` method and re-use it?

tests/Functional/SearchTest.php
```diff
@@ -978,6 +978,64 @@ public function testFilteringAndSortingForIdentifier(): void
...
+        ];
+        $athen = [
+            'lat' => 37.9838,
+            'lng' => 23.7275,
+        ];
+
+        $searchParameters = SearchParameters::create()
+            ->withFilter(sprintf(
+                '_geoBoundingBox(location, %s, %s, %s, %s)',
+                // Top Right (North East) (why north east see: https://github.com/loupe-php/loupe/issues/83)
```
Toflar, line 1003: I don't think we need the comments here. I think this can be explained in the docs abou the new `_geoBoundingBox()` filter which need to be added :)

tests/Unit/Internal/Filter/ParserTest.php
```diff
@@ -173,6 +173,19 @@ public static function filterProvider(): \Generator
             ],
         ];
 
+        yield 'Basic geo bounding box' => [
```
Toflar, line 176: Can we also add some tests in the `invalidFilterProvider()` just to make sure that invalid formats don't work? See other review about the coordinates.

src/Internal/Filter/Ast/GeoBoundingBox.php
```diff
@@ -0,0 +1,40 @@
...
+    ) {
+    }
+
+    public function getBbox(): Bounds
+    {
+        // phpgeo bounds are top left to bottom right but meilisearch and so loupe is top right to bottom left
+        return new Bounds(
```
Toflar, line 24: Maybe we could create the object in the constructor so that if you provide invalid coordinates, the `Parser` would already correctly fail?

Toflar (commented):
Awesome! You have implemented almost all that's needed, great!!

src/Internal/Filter/Ast/GeoBoundingBox.php
alexander-schranz: @Toflar the properties (north, east, ...) are currently `public` and could be changed from outside so we can not create the bound in constructor.

Toflar: There's no reason for them to be public?

alexander-schranz: orientated by existing GeoDistance, Filter classes

Toflar: Doesn't matter really, we should just make sure to catch invalid values on parsing :)

src/Internal/Search/Searcher.php
alexander-schranz: moved to an own method.

tests/Unit/Internal/Filter/ParserTest.php
alexander-schranz: added new tests.

tests/Unit/Internal/Filter/ParserTest.php
```diff
@@ -357,6 +370,36 @@ public static function invalidFilterProvider(): \Generator
...
+        yield 'Missing ) for _geoBoundingBox' => [
+            '_geoBoundingBox(location, 1.00, 2.00, 2.00, 3.00',
+            "Col 44: Error: Expected ')', got end of string.",
+        ];
+
+        yield 'Missing comma for _geoBoundingBox' => [
+            '_geoBoundingBox(location, 1.00 2.00, 2.00, 3.00)',
+            "Col 31: Error: Expected ',', got '2.00'",
+        ];
+
```
Toflar, line 402: Just one more test for invalid coordinates would be great :)

Toflar (commented):
Just one more test I think would be very helpful.
This looks very cool - how do you like it? :)

Toflar:
Oh and some docs on how to use the filter https://github.com/loupe-php/loupe/blob/main/docs/searching.md#filter

docs/searching.md
```diff
@@ -282,4 +282,52 @@ $results = [
...
\ No newline at end of file
+[Restaurant_Fixture]: ./../tests/Functional/IndexData/restaurants.json
+
+Additional to a query based on distance we can also search for locations inside a bounding box.
+In this example we have 4 documents and 3 with geo coordinates (New York, London, Vienna).
+We have a bounding context which goes from Dublin to Athen which so returns only London and Vienna.
+
+Keep in mind that the order of the arguments is important.
+The `_geoBoundingBox` expect `attributeName`, `north` (top), `east` (right), `south` (bottom), `west` (left).
+In this specific example top is Dublin Latitude, right is Athen Longitude, bottom is Athen Latitude and left is Dublin Longitude.
```
Toflar, line 293:
```suggestion
In this specific example, `top` is the latitude of Dublin,`right` is the longitude of Athens,`bottom` is the latitude of Athens and`left` equals Dublin's longitude.
```

docs/searching.md
```diff
@@ -282,4 +282,52 @@ $results = [
...
+
+```php
+$searchParameters = SearchParameters::create()
+    ->withAttributesToRetrieve(['id', 'name', 'location'])
+    ->withFilter('_geoBoundingBox(location, 53.3498, 23.7275, 37.9838, -6.2603)')
+;
+```
+
+This will result looks like this::
```
Toflar, line 302:
```suggestion
This is going to be your result:
```

Toflar (commented):
Some docs improvements and then I think this is ready to be merged. Awesome work, Alex!

Toflar:
Thanks @alexander-schranz! This is very cool! I'm going to clean up some stuff and then release it as v0.7 :)

alexander-schranz:
@Toflar Awesome. Thank you!

PR #93, opened 2024-11-04
Added support for Doctrine DBAL 4 and got rid of voku/portable-utf8

Toflar:


discordier:
Little problem here, the latest doctrine dbal is `3.9.3` but you added `3.10` here which is not released yet.

I currently work around this with: `"doctrine/dbal": "3.9.3 as 3.10.0",`

PR #98, opened 2024-11-20
Implemented a new ranking method (dropped TF-IDF)

Toflar:
Dropping TF-IDF in favor of a different ranking logic. Hopefully generating better search results 😊 
Currently we use

```
        /**
         * 1st: Number of query terms that match in a document.
         * 2nd: Proximity of the words
         */
```

Also they are weighted so we can adjust that if needed.


Toflar:
Merged this because the search results are already waaaay better than on the current `main` branch :) Feel free to test `develop` (there's also support for filtering for a `$searchParamters->withRankingScoreThreshold()` now 😎

PR #99, opened 2024-11-20
Restore doctrine dbal 3.x compatibility

discordier:
- Improve server version detection.
- Allow 3.9 again as 3.10 is not released yet.

Both broke in #93.


Toflar:
I thought `--prefer-lowest` would cover that use case but apparently Composer does not crash for non-existent versions 🙈 So It loaded 4.0.0 without me noticing. Thanks for fixing, Chris! Tag coming shortly.

PR #102, opened 2024-11-21
Additional SQLite connection optimizations

daun:
(Hi 👋 Fantastic piece of software you've created, thanks a lot for open-sourcing it!)

I've been toying a bit with optimizing sqlite connections on production sites and thought these might make sense to include by default. The good thing is that Loupe is usually the only client of its particular database, so it can make a few assumptions about how to optimize the connection. Not sure if you agree with all of these suggestions, but it'd be great to get a discussion started.

- Add a few optimizations to reduce disk I/O
- Add a timeout to avoid errors if db is locked by another instance
- Enable incremental vacuum to auto-reduce database size
- Tested this on a few sites and I'm seeing about a 10% increase in speed (as measured by `processingTimeMs`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Note that some implementation details (e.g. libraries used) referenced in this b
 Performance depends on many factors but here are some ballpark numbers based on indexing the [~32k movies fixture by 
 MeiliSearch][MeiliSearch_Movies] and the test files in `bin` of this repository:
 
-* Indexing (`php bin/index_performance_test.php`) will take a little over 3min (~170 documents per second)
+* Indexing (`php bin/index_performance_test.php`) will take a little over 2min (~230 documents per second)
 * Querying (`php bin/search_performance_test.php`) for `Amakin Dkywalker` with typo tolerance enabled and ordered by 
   relevance finishes in about `120 ms`
 
--- a/src/LoupeFactory.php
+++ b/src/LoupeFactory.php
@@ -87,8 +87,7 @@ final class LoupeFactory
             ));
         }
 
-        // Use Write-Ahead Logging if possible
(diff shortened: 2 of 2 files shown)


Toflar:
Hey man, thanks a lot for the prop, the PR and taking the time to contribute - love it!
Would you mind targeting this against `develop` (future version `0.8`), please?
Regarding the bench: I've included 2 scripts in the `bin` folder, can you share your results pre and post PR? 😎

Toflar:
FYI: Indexing will take about 3 minutes :) You can adjust the `config.php` in the `bin` directory to use the `movies_1000.json` for quicker local tests before running the big one :)

daun:
@Toflar Sure! Will do the benchmarks when I find a minute.

Toflar:
I will bench on my machine too and let you know. I've just created another PR that improved indexing for me, would love to have your feedback on that too 😊 https://github.com/loupe-php/loupe/pull/103

Toflar:
Oh man, the boost on my system is huge!! https://github.com/loupe-php/loupe/pull/103 brings the time from about 220s to 188s and together with your PR we are now at 137s!

daun:
Sounds fantastic! I'm having trouble getting the benchmarks to run, but it looks like the results on your machine are speaking for themselves :)

Toflar:
Sure, please just rebase on `develop`, then I'll happily merge it :-)

daun:
This might hurt performance when reading/searching, though, that's why I've been a bit conservative with page size.

Toflar:
Does not have any effect for me :)

Toflar:
Thanks a lot Philipp!

PR #104, opened 2024-11-22
Negated search queries

### LINKED ISSUE #95: Add support for negative queries
- Support `-foobar` to exclude documents that contain `foobar`
- Support `-"foo bar"` to exclude documents that contain the entire `"foo bar"` string

daun:
This was way easier than I thought 🎉 Very eager to have you give this a spin.

- Implement negated queries and query phrases: `-lorem` and `-"lorem ipsum"`
- Adjust the tokenizer to mark tokens and phrases as negated
- Add tests for negated tokens and negated phrases
- Haven't touched the parser — that's only used for filtering, if I understand correctly?
- ~~I've decided to skip stemming on negated terms, let me know what you think about that part~~
- Closes #95


Toflar:
Nice!! Congratulations on your first deep-dive 😎 Yes, the parser is only for filtering, that's true - so your code here looks just fine!
I am not sure the AND-OR grouping with negated terms is correct though. Let's look at a longer phrase like `compact round "red car" coating` (total nonsense, I know).
This needs to translate (and currently does) into:

`compact OR round OR (red AND car) OR coating`

With negated terms, this now becomes more complex: `compact -round "red car" coating` must be

`compact OR coating OR (red AND car) AND (NOT round)`

I am not quite sure the logic is currently correct - I might be wrong though but we might need some test with a longer query to double check the SQL that's built is correct? What do you think?

> I've decided to skip stemming on negated terms, let me know what you think about that part

I honestly don't know how other search engines do that. I'm trying to find out for Meilisearch.

Toflar:
> I honestly don't know how other search engines do that. I'm trying to find out for Meilisearch.

I'm fairly sure MS applies stemming to both, regular as well as negated terms unless they are part of a phrase. Which I think we should do in Loupe too. Wdyt?

Toflar:
Like if you go for `-learning` you would also not get any documents with `learn` which I think is correct. If you want to exclude only `learning` explicitly, this is doable using `-"learning"`, right?

daun:
I'm happy to be convinced on both points. Let's take our time and test this with more complex queries. Stemming everywhere also sounds reasonable, especially given the option of excluding whole phrases as you mentioned.

daun:
And should the query be grouped to correctly handle precedence of and/or here?

`(compact OR coating OR (red AND car)) AND (NOT round)`

Toflar:
I don't think it's necessary to keep the order of the terms like they are in query. The important thing is that all terms are considered as `OR` except for phrases and negated expressions. Phrases have to generate an `AND` within the phrase and negated terms become a top level `AND`.

daun:
In my head, the queries should be correct, then. It first groups all positive terms, joined by `OR` and wrapped in parenthesis. It then appends all negated terms joined by `AND`. Let me double check.

Toflar:
Right now we keep the order. So `compact round "red car" coating` becomes `compact OR round OR (red AND car) OR coating`. But it would be totally okay to go `compact OR round OR coating OR (red AND car)`.



> In my head, the queries should be correct, then. It first groups all positive terms, joined by `OR` and wrapped in parenthesis. It then appends all negated terms joined by `AND`. Let me double check.

I agree, I just want to have a test that maybe has a longer query, jumping between phrases and negated stuff, just to make sure our grouping is correct :)

daun:
So this is what I'm getting for a search of `friendly mother -boy -"depressed suburban father" father`.

The resulting SQL definitely looks wrong to me, and the results also aren't exactly what I'm expecting. Will investigate!

```sql
SELECT COUNT() OVER() AS totalHits,
       d.document
FROM documents d
WHERE (((d.id IN
           (SELECT DISTINCT document
            FROM _cte_term_document_matches_0_friendly))
        OR (d.id NOT IN
              (SELECT DISTINCT document
               FROM _cte_term_document_matches_3_depressed)
            AND d.id NOT IN
              (SELECT DISTINCT document
               FROM _cte_term_document_matches_4_suburban)
            AND d.id NOT IN
              (SELECT DISTINCT document
               FROM _cte_term_document_matches_5_father))
        OR (d.id IN
              (SELECT DISTINCT document
               FROM _cte_term_document_matches_6_father))))
  AND (((d.id IN
           (SELECT DISTINCT document
            FROM _cte_term_document_matches_1_mother))
        AND (d.id NOT IN
               (SELECT DISTINCT document
                FROM _cte_term_document_matches_2_boy))))
```

daun:
This is what it comes down to. Definitely off.

```
(
        friendly
    OR
    (
            NOT depressed
        AND
            NOT suburban
        AND
            NOT father
    )
    OR
        father
)
AND
(
        mother
    AND
        NOT boy
)
```

src/Internal/Search/Searcher.php
```diff
@@ -148,7 +148,7 @@ public function fetchResult(): SearchResult
 
     public function getCTENameForToken(string $prefix, Token $token): string
     {
-        return $prefix . $token->getId();
+        return $prefix . $token->getId() . '_' . $token->getTerm();
```
Toflar, line 151: Be aware that this might break the queries. Not sure what happens if you use special characters in your term ;-)

daun: Good point! We can a) revert this once we're happy with the result or b) sanitize the term. I like the debuggability of including the term, but it's also not really required.

daun:
Got the grouping working. There's some other tests that are now broken, I'm marking the PR as a draft for the time being.

Does this look okay to you? It's looking fine to me and fetching the results I'm expecting.

`friendly mother -boy -"depressed suburban father" father`

```
SELECT COUNT() OVER() AS totalHits,
       d.document
FROM documents d
WHERE (((d.id IN
           (SELECT DISTINCT document
            FROM _cte_term_document_matches_0_friendly))
        OR (d.id IN
              (SELECT DISTINCT document
               FROM _cte_term_document_matches_1_mother))
        OR (d.id IN
              (SELECT DISTINCT document
               FROM _cte_term_document_matches_6_father))))
  AND (((d.id NOT IN
           (SELECT DISTINCT document
            FROM _cte_term_document_matches_2_boy))
        AND (d.id NOT IN
               (SELECT DISTINCT document
                FROM _cte_term_document_matches_3_depressed)
             AND d.id NOT IN
               (SELECT DISTINCT document
                FROM _cte_term_document_matches_4_suburban)
             AND d.id NOT IN
               (SELECT DISTINCT document
                FROM _cte_term_document_matches_5_father))))
```

src/Internal/Search/Searcher.php
daun: Removed it for now, and left it as a comment to aid future debugging efforts :)

daun:
I've tried the `Group` helper route, but found it wasn't really necessary. Grouping the `TokenCollection` into phrases proved to be somewhat helpful, so I implemented it that way for now.

Toflar:
I had 0 additional remarks. Coding style was perfect and nicely commented! Nothing more to say than thank you for this awesome contribution! ❤️ 
I'll likely publish v0.8 next week, looking forward to it 😎

daun:
Great, looking forward to a release! 🧃

daun:
@discordier That's very probably caused by this, then. I'll add some tests to see if we need to tweak things a bit.

PR #107, opened 2024-11-22
Stop words

### LINKED ISSUE #61: Stop words
Are stop words supported? If yes, how do I add them?

daun:
- Add search param setting for stop words to ignore in queries
- Improves query performance slightly when lots of generic words are included
- In case of stopword-only queries: fall back to all words, i.e. act like there were no stop words
- Closes #61

## Usage

```php
$configuration = \Loupe\Loupe\Configuration::create()
    ->withStopWords(['a', 'by', 'the']);
```

## Questions

- Could/should this be combined with automatic language detection? I've found lots of stop word lists online, but many of them are extremely opinionated and I wouldn't want those defaults to be present on my sites. So I'd prefer explicitly passing in stopwords at query time, but you might have a different preference here :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Loupe…
 * …supports negative keyword and phrase search using `-` as modifier
 * …supports filtering (and ordering) on any attribute with any SQL-inspired filter statement
 * …supports filtering (and ordering) on Geo distance
-* …orders relevance based on a number of factors such as nubmer of matching terms as well as proximity
+* …orders relevance based on a number of factors such as number of matching terms as well as proximity
 * …auto-detects languages
 * …supports stemming
 * …is very easy to use
@@ -62,13 +62,11 @@ I even took the liberty to copy some of their test data to feed Loupe for functi
 
 ## Usage
 
+### Creating a client
+
+The first step is configuring and creating a client.
+
 ```php
-<?php
-
-namespace App;
-
-require_once 'vendor/autoload.php';
-
 use Loupe\Loupe\Config\TypoTolerance;
 use Loupe\Loupe\Configuration;
 use Loupe\Loupe\LoupeFactory;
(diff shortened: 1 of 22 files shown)


Toflar:
Hey, thanks for working on this! I think, stop words should not be part of the `SearchParameters` but instead of `Configuration`. What if you have a list of 1000 stop words, would you pass them to Loupe on every single search request? 😊 Moreover, it doesn't make sense to even index terms that are part of the stop words.
Imho, we should go similar to https://www.meilisearch.com/docs/reference/api/settings#get-stop-words so `$engine->addStopWords()`, `$engine->getStopWords()` and `$engine->deleteStopWords()`. What do you think?

Providing lists of common stop words is a separate question and I would ignore this for now.

daun:
Very good points! For more than a few stop words, it'd make sense to store them in the database. However I'm not sure where you'd even get more than 40 or 50 stop words from. Setting stop words for all languages probably won't make sense since one language's stop words are another language's keywords ;)

This is Elastic's default list of English stop words, 33 in number, and half of those seem a bit too aggressive to me: a, an, and, are, as, at, be, but, by, for, if, in, into, is, it, no, not, of, on, or, such, that, the, their, then, there, these, they, this, to, was, will, with.

If possible, I like the idea of keeping it simple for now and putting them into the configuration. If long-term conceptual compatibility with Meilisearch is a goal (which makes perfect sense), let's leave this open for now and see when I get to moving stop words into the engine/database.

daun:
Let's go with the Meilisearch implementation then and store things from the engine. The willingness to run with good decisions made by others is what made loupe interesting in the first place. I won't get to this for a few days, though, let's just keep an open tab on it for now.

daun:
@Toflar I've moved the stop words into the engine. Good to go from my end :) I've opted not to store them in the database permanently. The complexity of storing and retrieving them from a table didn't seem to make much sense when only dealing with 10-50 words that can be passed in during configuration. Guess we can always add that later if it makes sense? But I can't see someone using 1000s of stop words realistically.

Toflar:
Yeah, what we're doing then is externalizing the responsibility of storing them. If e.g. you want to give some user the ability to configure the stop words themselves, you'll have to store them somewhere and pass them to the `Configuration` everytime you instantiate Loupe. By storing them within Loupe, you could potentially omit that.
But there are a lot of other things you potentially want to have users to be able to configure which are not stored in the index itself so I guess we're good for the moment. We can still enhance the feature (or whoever needs it can 😊 ).

Toflar:
Thanks @daun!

daun:
@Toflar Sounds good!

PR #108, opened 2024-11-22
Client pool

### LINKED ISSUE #15: Add a LoupManager to manage multipe Indexes and save configurations
Save the configuration can let the search engine run with the data it was created it and make the changing the configuration a more active method call can even allow to have 2 instances of a Index a Read with old configuration and a Write with the new configuration.

This could be achieved by a LoupeManager which looks like this:

```php
class LoupeManager
{
    /**
     * @var Loupe[]
     */
    private array $loupe = [];

    public function __construct(
        private readonly LoupeFactory $loupeFactory,
        private string $directory,
    ) {
    }
    
    public function getLoupe(string $index): Loupe
    {
        if (!isset($this->loupe[$index->name])) {
            $configurationFile = $this->directory . $index . '.loup';

            if (!\file_exists($configurationFile)) {
                throw new \LogicException('Index need to be created before accessing Loupe instance.');
            }

            /** @var string $configurationContent */
            $configurationContent = \file_get_contents($configurationFile);

            /** @var Configuration $configuration */
            $configuration = \unserialize($configurationContent);
            
            $this->loupe[$index->name] = $this->loupeFactory->create($this->directory . $index . '.db', $configuration);
        }

        return $this->loupe[$index->name];
    }
    
    public function existIndex(string $index): bool
    {
        return file_exists(($this->directory . $index . '.db', $this->directory . $alias . '.db');
    }
    
    public function dropIndex(string $index): void
    {
         unlink($this->directory . $index . '.db', $this->directory . $alias . '.db');
         unlink($this->directory . $index . '.loup', $this->directory . $alias . '.loup');
    }
    
    public function createIndex(string $index, Configuration $configuration): void
    {
        $configurationContent = \serialize($configuration);
        
        file_put_contents($this->directory . $index . '.loup', $this->directory . $alias . '.loup', $configurationContent);
        touch($this->directory . $index . '.db');
    }
    
    public function alias(string $index, string $alias): void
    {
         // aliases via symlinks for zerodowntime reindexing
         symlink($this->directory . $index . '.db', $this->directory . $alias . '.db');
         symlink($this->directory . $index . '.loup', $this->directory . $alias . '.loup');
    }
    
    public function removeAlias(string $alias): void
    {
         unlink($this->directory . $alias . '.db');
         unlink($this->directory . $alias . '.loup');
    }
}
```

```php
$readLoupe = $loupeManager->getLoupe('blog');

$loupeManager->createIndex('blog_202208050332', $newConfiguration);
$writeLoupe = $loupeManager->getLoupe('blog_202208050332');

$writeLoupe->addDocuments([/* ... */]);

$loupeManager->alias('blog', 'blog_202208050332');
```

daun:
- Create a client pool to manage multiple indexes
- Allow creating/deleting/checking existence of indexes by name
- Client pool has a base path that contains all indexes
- Move `symfony/filesystem` into normal dependencies
- Called this `Manager` originally, but `ClientPool` sounds like something at home in the database world
- Very open to renaming this, though
- Closes #15

## Usage

```php
$pool = new LoupeClientPool('/storage/search/');
$config = Configuration::create()->withPrimaryKey('uid');

// Create clients for separate indexes
$movieIndex = $pool->get('movies', $configuration);
$docIndex = $pool->get('documents', $configuration);

// Check if index exists
$pool->indexExists('documents');

// Drop index
$pool->dropIndex('documents');
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,11 +29,11 @@
         "mjaschen/phpgeo": "^4.2",
         "toflar/state-set-index": "^2.0.1",
         "psr/log": "^2.0 || ^3.0",
-        "nitotm/efficient-language-detector": "^2.0"
+        "nitotm/efficient-language-detector": "^2.0",
+        "symfony/filesystem": "^6.2"
     },
     "require-dev": {
         "symfony/var-dumper": "^6.2",
-        "symfony/filesystem": "^6.2",
         "symplify/easy-coding-standard": "11.2.4.72",
         "symfony/finder": "^6.2",
         "phpunit/phpunit": "^10.0",
--- a/src/LoupeClientPool.php
+++ b/src/LoupeClientPool.php
@@ -1,10 +1,9 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Loupe\Loupe;
 
-use Loupe\Loupe\Configuration;
-use Loupe\Loupe\Loupe;
-use Loupe\Loupe\LoupeFactory;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Filesystem\Path;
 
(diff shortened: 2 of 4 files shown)


daun:
I reconsidered this, and I don't think this is the right abstraction. Deleting and recreating the SQLite file to drop an index causes ongoing file permission issues on live servers. A better method would be to drop all tables and rows while keeping the SQLite file, as this [zap proposal](https://github.com/loupe-php/loupe/discussions/87) suggested. This way, users can set the file permissions once when they set up a project, and not have to re-adjust them regularly.

PR #109, opened 2024-11-22
Usage docs

daun:
- Slight streamlining of the usage docs
- Split examples into sections to make things more approachable
  - Creating a client
  - Adding documents
  - Performing search
- Remove vendor/autoload include and namespaces to simplify things


Toflar:
Thank you!

PR #110, opened 2024-11-23
Delete all documents

daun:
- Add a method to delete all documents
- Useful when creating an index from scratch in a schedule or cli command
- A previous attempt of deleting and recreating the sqlite file resulted in nasty file permissions issues
- This approach works much better and feels less hacky
- Tests are added

### DIFF
--- a/src/Internal/Index/Indexer.php
+++ b/src/Internal/Index/Indexer.php
@@ -95,6 +95,20 @@ class Indexer
         return new IndexResult($successfulCount, $documentExceptions);
     }
 
+    public function deleteAllDocuments(): self
+    {
+        if ($this->engine->getIndexInfo()->needsSetup()) {
+            return $this;
+        }
+
+        $this->engine->getConnection()
+            ->executeStatement(sprintf('DELETE FROM %s', IndexInfo::TABLE_NAME_DOCUMENTS));
+
+        $this->reviseStorage();
+
+        return $this;
+    }
+
     /**
      * @param array<int|string> $ids
      */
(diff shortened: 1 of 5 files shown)


src/Internal/Index/Indexer.php
```diff
@@ -95,6 +95,20 @@ public function addDocuments(array $documents): IndexResult
...
+    public function deleteAllDocuments(): self
+    {
+        if ($this->engine->getIndexInfo()->needsSetup()) {
+            return $this;
+        }
+
+        $this->engine->getConnection()
+            ->executeStatement(sprintf('DELETE FROM %s', IndexInfo::TABLE_NAME_DOCUMENTS));
+
+        $this->reviseStorage();
```
Toflar, line 107: I just noticed that `reviseStorage()` does not handle the `state_set` table. We should delete all states that are not part of `terms` and `prefixes_terms` anymore. Otherwise deleting all documents will cause the stateset to contain false entries.

daun: Let me look into that :)

Toflar:
You don't have to understand what it's doing, really. The typo tolerance stuff is most likely the most complex part about Loupe. I read that paper a gazillion times 🤣  Imho all we have to here is to DELETE all `state`s in `state_set` that are not referenced in either `terms.state` or `prefixes_terms.state`, that's it. But I can also look into it - feel free to create an issue if you don't want to take a look at it - then I will 👍

daun:
Sounds good! I've created #112 to track this.

Toflar:
Thanks! Then the only thing left to do here is to document that new method in the docs 😊

Toflar (approved):

Toflar:
Thank you, @daun ❤️

discordier:
I think this is the implementation of  our sept discussion https://github.com/loupe-php/loupe/discussions/87

As I read it, currently, we now have false positives until the state set gets cleaned up?

Toflar:
> As I read it, currently, we now have false positives until the state set gets cleaned up?

No, I don't think that's the case but I will have to double check.

PR #111, opened 2024-11-24
Attribute ranking order

daun:
(Sorry for spamming you with pull requests, but it's just too much fun 🤠)

Add a new relevance factor: attribute ranking order. Give slight precedence to matches that appear early in the list of searchable attributes. This will naturally rank matches in the `title` higher, assuming `title` appears first in the list.

The behavior is modeled on Meilisearch. When searchable attributes are not configured (`['*']`), attribute order is ignored. When they are configured, they are ranked in order of appearance: `['title', 'summary']` will slightly prefer `title` matches. The factor is currently `0.8`, I'm seeing reasonable results with that. The factor below `1` technically downranks matches in later attributes, rather than uprank earlier attributes, but this way we don't get final relevance scores above `1`.

See the Meilisearch docs on [Attribute ranking order](https://www.meilisearch.com/docs/learn/relevancy/attribute_ranking_order) for details.

### No intrinsic ranking

```php
Configuration::create()->withSearchableAttributes(['*'])
```

### Intrinsic ranking by order

```php
# Prefer matches by order of appearance in this list: `title` > `summary` > `body`

Configuration::create()->withSearchableAttributes(['title', 'summary', 'body'])
```

## Example matches

Example rankings when searching for **`game of life`**. The first relevance column uses `['*']` as searchable attributes, the second relevance column uses `['title', 'content']`

| Title | Content | Without attribute order | With attribute order |
|-------|---------|----------------|-------------|
| **Game of life** | A thing with everlasting learning | 1.0 | 1.0 |
| Everlasting learning | The unexamined **game of life** | 1.0 | 0.88 |
| Learning to **game** | What **life** taught me about learning | 0.83 | 0.78 |

### DIFF
--- a/src/Internal/Search/Sorting/Relevance.php
+++ b/src/Internal/Search/Sorting/Relevance.php
@@ -129,11 +129,11 @@ class Relevance extends AbstractSorter
             []
         );
 
-        $matchedAttributeWeights = array_map(fn ($attribute) => $attributeWeights[$attribute] ?? null, array_unique($matchedAttributes));
+        $matchedAttributeWeights = array_map(fn ($attribute) => $attributeWeights[$attribute] ?? 1, $matchedAttributes);
         $matchedAttributeWeights = array_filter($matchedAttributeWeights, fn ($weight) => $weight !== 1);
 
         return count($matchedAttributeWeights) ?
-            (array_sum($matchedAttributeWeights) / count($matchedAttributeWeights))
+            (array_sum($matchedAttributeWeights) / count($positionsPerTerm))
             : 1;
     }
 
--- a/src/Internal/Search/Sorting/Relevance.php
+++ b/src/Internal/Search/Sorting/Relevance.php
@@ -39,7 +39,7 @@ class Relevance extends AbstractSorter
             return;
         }
 
-        $weights = $searcher->getSearchParameters()->getAttributeWeights();
+        $weights = $this->calculateIntrinsicAttributeWeights($engine);
 
         $select = sprintf(
             "loupe_relevance((SELECT group_concat(%s, ';') FROM (%s)), %s, '%s') AS %s",
@@ -81,6 +81,21 @@ class Relevance extends AbstractSorter
         return \count($matchedAttributeWeights) ?
             (array_sum($matchedAttributeWeights) / \count($positionsPerTerm))
             : 1;
+    }
+
+    /**
+     * @return array<string, int>
+     */
+    public static function calculateIntrinsicAttributeWeights(Engine $engine): array
+    {
+        $searchableAttributes = $engine->getConfiguration()->getSearchableAttributes();
+        if ($searchableAttributes === ['*']) {
+            return [];
+        }
+
+        // Assign linear weight to each attribute that is searchable
+        // ['title', 'summary', 'body] → ['title' => 3, 'summary' => 2, 'body' => 1]
+        return array_combine($searchableAttributes, range(count($searchableAttributes), 1, -1));
     }
 
     /**
@@ -193,7 +208,7 @@ class Relevance extends AbstractSorter
     /**
      * Parse an intermediate string representation of attribute weights back into an array
      *
-     * "title:3;summary:4" -> ["title" => 3, "summary" => 4]
+     * "title:0;summary:1" -> ["title" => 0, "summary" => 1]
      *
      * @return array<string, int>
      */
--- a/src/Internal/Search/Sorting/Relevance.php
+++ b/src/Internal/Search/Sorting/Relevance.php
@@ -11,7 +11,7 @@ use Loupe\Loupe\Internal\Search\Searcher;
 class Relevance extends AbstractSorter
 {
     /**
-     * @var array<string, array<string, float>> $attributeWeightValuesCache
+     * @var array<string, array<string, float>>
      */
     protected static array $attributeWeightValuesCache = [];
 
--- a/src/Internal/Search/Sorting/Relevance.php
+++ b/src/Internal/Search/Sorting/Relevance.php
@@ -52,13 +52,14 @@ class Relevance extends AbstractSorter
 
         $searcher->getQueryBuilder()->addSelect($select);
 
-        // No need to use the abstract addOrderBy() here because the relevance alias cannot be of our internal null or empty
-        // value
+        // No need to use the abstract addOrderBy() here because the relevance alias cannot be of
+        // our internal null or empty value
         $searcher->getQueryBuilder()->addOrderBy(Searcher::RELEVANCE_ALIAS, $this->direction->getSQL());
 
         // Apply threshold
-        if ($searcher->getSearchParameters()->getRankingScoreThreshold() > 0) {
-            $searcher->getQueryBuilder()->andWhere(Searcher::RELEVANCE_ALIAS . '>= ' . $searcher->getSearchParameters()->getRankingScoreThreshold());
+        $threshold = $searcher->getSearchParameters()->getRankingScoreThreshold();
+        if ($threshold > 0) {
+            $searcher->getQueryBuilder()->andWhere(Searcher::RELEVANCE_ALIAS . '>= ' . $threshold);
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/Internal/Search/Sorting/Relevance.php
```diff
@@ -39,28 +40,88 @@ public function apply(Searcher $searcher, Engine $engine): void
             return;
         }
 
+        $searchableAttributes = $engine->getConfiguration()->getSearchableAttributes();
+        $weights = $this->calculateIntrinsicAttributeWeights($searchableAttributes);
```
Toflar, line 44: Should be called statically?

src/Internal/Search/Sorting/Relevance.php
```diff
@@ -39,28 +40,88 @@ public function apply(Searcher $searcher, Engine $engine): void
...
+    public static function calculateIntrinsicAttributeWeights(array $searchableAttributes): array
+    {
+        if ($searchableAttributes === ['*']) {
+            return [];
+        }
+
+        // Assign decreasing weights to each attribute
+        // ['title', 'summary', 'body] → ['title' => 1, 'summary' => 0.8, 'body' => 0.8 ^ 2]
+        $weight = 1;
+        $factor = Configuration::ATTRIBUTE_RANKING_ORDER_FACTOR;
```
Toflar, line 98: Can inline `$factor` and save those variable assignments?

src/Internal/Search/Sorting/Relevance.php
```diff
@@ -70,14 +131,19 @@ public static function calculateProximityFactor(array $positionsPerTerm, float $
         $positionPrev = null;
 
         foreach ($positionsPerTerm as $positions) {
+            // Account for old format without attribute: [1,2,3;4,5] → [[1,null],[2,null],[3,null];[4,null],[5,null]]
+            $positions = array_map(fn ($position) => !\is_array($position) ? [$position, null] : $position, $positions);
```
Toflar, line 135: You don't have to do that, do you? It's happening during search so it's not something that was persistent before and we may assume that the query includes this info now?

src/Internal/Search/Sorting/Relevance.php
```diff
@@ -98,43 +164,44 @@ public static function calculateProximityFactor(array $positionsPerTerm, float $
...
         $totalQueryTokenCount = (int) $totalQueryTokenCount;
-        $positionsPerTerm = array_map(fn ($term) => array_map('intval', explode(',', $term)), explode(';', $positionsInDocumentPerTerm));
+        $positionsPerTerm = static::parseTermPositions($positionsInDocumentPerTerm);
+        $attributeWeightValues = static::parseAttributeWeights($attributeWeights);
+
+        // Higher weight means more importance
+        $relevanceWeights = [
+            2, // 1st: Number of query terms that match in a document
+            1, // 2nd: Proximity of the words
+            1, // 3rd: Weight of attributes matched
```
Toflar, line 185:
```suggestion
            1, // 3rd: Weight of attributes matched (use 1 as they are already weighted)
```

tests/Functional/SearchTest.php
```diff
@@ -1832,7 +1832,7 @@ public function testRankingWithLotsOfMatches(): void
                 [
                     'id' => 22,
                     'title' => 'Pirates of the Caribbean: The Curse of the Black Pearl',
-                    '_rankingScore' => 1.0,
+                    '_rankingScore' => 0.79194,
```
Toflar, line 1835: I don't think this is correct? The searchable fields are configured like this: `->withSearchableAttributes(['title', 'overview'])` in that test. So in my opinion, this is an exact match in the `title`, so the most important, attribute. This should result in `1.0` - don't you think?

src/Internal/Search/Sorting/Relevance.php
```diff
@@ -98,43 +164,44 @@ public static function calculateProximityFactor(array $positionsPerTerm, float $
...
-         * 2nd: Proximity of the words
-         */
-        static $relevanceWeights = [2, 1]; // Higher weight means more importance
-        static $totalWeight = 3; // Must be the sum of the above
-
-        $relevanceFactors = [];
         $totalQueryTokenCount = (int) $totalQueryTokenCount;
-        $positionsPerTerm = array_map(fn ($term) => array_map('intval', explode(',', $term)), explode(';', $positionsInDocumentPerTerm));
+        $positionsPerTerm = static::parseTermPositions($positionsInDocumentPerTerm);
+        $attributeWeightValues = static::parseAttributeWeights($attributeWeights);
```
Toflar, line 179: This is done on every document in the search result. But it's always the same, no? We can save quite a bit of performance here maybe?

Toflar (commented):
> (Sorry for spamming you with pull requests, but it's just too much fun 🤠)

No worries, I love your work! ❤️ I felt the same when I created Loupe 😉 

I would've modelled this exactly after MeiliSearch as well and it would've looked more or less exactly the same so I'm super excited and positive about this PR! 🥳 
I've reviewed and have some comments on the implementation, let's see where this goes!

src/Internal/Search/Sorting/Relevance.php
daun: Fixed :)

src/Internal/Search/Sorting/Relevance.php
daun: Inlined :) Makes sense to optimize for memory and speed here.

src/Internal/Search/Sorting/Relevance.php
daun: I was trying to avoid updating the existing tests 🤡 Updated the tests and this is no longer required.

src/Internal/Search/Sorting/Relevance.php
daun: Great point, added a cache for those.

src/Internal/Search/Sorting/Relevance.php
```diff
@@ -39,28 +44,87 @@ public function apply(Searcher $searcher, Engine $engine): void
...
         }
+
+        // Assign decreasing weights to each attribute
+        // ['title', 'summary', 'body] → ['title' => 1, 'summary' => 0.8, 'body' => 0.8 ^ 2]
+        $weight = 1;
+        return array_reduce(
+            $searchableAttributes,
+            function ($result, $attribute) use (&$weight) {
+                $result[$attribute] = round($weight, 2);
+                $weight *= 0.8;
```
Toflar, line 106: Why did you drop the reference to `Configuration::ATTRIBUTE_RANKING_ORDER_FACTOR`? It's a constant, I meant we can just reference it right away 😊

tests/Functional/SearchTest.php
daun: This one's tricky — some of those search terms also appear in the `overview` which is ranked `0.8` since it's second in order. That's why the `0.8` is applied here, plus some rounding it seems. What's a better way of handling this? We might add some logic for checking if all terms appear in a single attribute and apply the higher value in that case?

daun: Yeah, that's probably what I'd expect when using the search...

src/Internal/Search/Sorting/Relevance.php
Toflar: It's a constant, OPcache will inline this anyway :) I commented on your `$factor` variable which made no sense to me :)

tests/Functional/SearchTest.php
Toflar: I see. We should maybe check what Meilisearch does but as a user, I would expect this to be `1.0`. It is an absolutely exact hit on the most important attribute. This is the defnition of `1.0`.

src/Internal/Search/Sorting/Relevance.php
daun: Fixed to use the static property again, but this time inlined.

tests/Functional/SearchTest.php
daun: I've updated it to prefer the higher factor in case a term matches across multiple attribute. Seems to solve this nicely :)

Toflar (approved):

Toflar:
Love it! ❤️ What do you think, should we add some info about the ranking algorithm to the docs? Previously, there was not much to document as you could not influence it anyway. Now you can by defining the order of the searchable attributes. So I think now it would deserve some docs...

daun:
@Toflar Sure! Let me whip up a quick draft.

daun:
@Toflar I've created a separate PR #113 with new docs on the ranking factors. Let's discuss over there :)

Toflar:
Thanks for adding a third ranking factor ❤️

PR #118, opened 2024-11-29
Update to toflar/state-set-index v3

Toflar:
Keeping the transposition distance at `2` for BC now. When we want to change to 1 - aka Damerau-Levenshtein - we also need to adjust other parts of Loupe.


Toflar:
@ausi - care to review this? 😇

ausi (approved):
LGTM 🎉

PR #120, opened 2024-11-30
Remove orphaned terms after updating documents

### LINKED ISSUE #119: Orphaned terms are not removed after updating a document
When updating an existing document via `$loupe->addDocument()`, new terms are added, but orphaned terms aren't removed. Searching for the old value still brings up the existing document, even though it no longer contains the term. The logic currently works after **deleting** documents, we just need a way to apply it after **updating** documents as well.

Simplified pseudo-code:

```php
$loupe->addDocument([
  'id' => 1
  'title' => 'Vienna',
]);

$loupe->search('Vienna'); // returns document #1

$loupe->addDocument([
  'id' => 1
  'title' => 'Munich',
]);

$loupe->search('Vienna'); // still returns document #1, should return nothing
```

I think the issue is that during indexing, all new terms are added without first removing existing terms, which makes it impossible for the `reviseStorage()` method to even know which ones are unused. Which offers the solution, as well: first remove all terms linked to a document, and only then add the terms.

Suggested fix at #118

Discovered while working on (and currently blocking) #117

daun:
- Fix an issue where updating a document won't remove orphaned terms from the index
- The suggested solution is rather crude: delete documents before (re-)indexing them
- There's probably a more elegant solution using upsert, but I can't think of one :)
- Made sure to disable storage revision when making those types of deletions
- Opted for single deletion instead of batching to play nice with transactions and allow rollback
- Fixes #119

### DIFF
--- a/src/Internal/Index/Indexer.php
+++ b/src/Internal/Index/Indexer.php
@@ -60,6 +60,10 @@ class Indexer
 
                             $this->engine->getConnection()
                                 ->transactional(function () use ($document) {
+                                    // Delete the document first if it already exists to avoid orphaned data
+                                    $userId = (string) $document[$this->engine->getConfiguration()->getPrimaryKey()];
+                                    $this->deleteDocuments([$userId], reviseStorage: false);
+
                                     $documentId = $this->indexDocument($document);
                                     $this->indexMultiAttributes($document, $documentId);
                                     $this->indexTerms($document, $documentId);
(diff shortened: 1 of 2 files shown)


Toflar (changes requested):
I would've done exactly the same 😊 I don't think this is a bad approach. But can you target that at `main` please because this affects 0.8 (aka current stable) as well so we should fix it there. I'll merge it into develop after.

daun:
@Toflar Rebased on main :) Can't see if the tests are passing. They do locally, anyway.

Toflar (approved):

PR #122, opened 2024-12-02
Switch to Damerau-Levenshtein

Toflar:
This PR would enable @ausi's new implementation on the Damerau-Levenshtein algorithm.
For those not familiar with it: Compared to regular Levenshtein, transpositions are counted as the cost of 1 typo as opposed to 2 typos in regular Levenshtein. What's a transposition? Scrambled letters right next to one another: e.g. "huose" instead of "house".

It happens quite often when users type. Hence I'd consider Damerau-Levenshtein to be more suitable for a search engine.
However, the PHP version is quite a bit slower than `levenshtein()` which is a native PHP/C function.

So currently, the benchmark in `bin` is about 20ms slower (140ms instead of 120ms). However, the typos are a bit more UX friendly...so I guess it's a trade-off here at the moment.

I was fiddling around with Damerau-Levenshtein automatons which might improve this a little but I'm not sure.

### DIFF
--- a/src/Internal/Levenshtein.php
+++ b/src/Internal/Levenshtein.php
@@ -4,11 +4,13 @@ declare(strict_types=1);
 
 namespace Loupe\Loupe\Internal;
 
+use Toflar\StateSetIndex\DamerauLevenshtein;
+
 class Levenshtein
 {
-    public static function levenshtein(string $string1, string $string2, bool $firstCharTypoCountsDouble): int
+    public static function damerauLevenshtein(string $string1, string $string2, bool $firstCharTypoCountsDouble): int
     {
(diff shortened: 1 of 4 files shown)


daun:
Most users are okay waiting 20ms longer if the results make more sense to them once they come in 🤠 And if every millisecond counts, I guess the recommendation would be to migrate to Meilisearch on a high end server anyway.

ausi:
> So currently, the benchmark in `bin` is about 20ms slower (140ms instead of 120ms).

With https://github.com/Toflar/state-set-index/pull/9 the benchmark got only slightly slower (on average) from 112ms vs 116ms on my machine.

Toflar:
I'm merging this even though https://github.com/Toflar/state-set-index/pull/9 is not released yet but I will make sure it is once Loupe 0.9 is published.

PR #124, opened 2024-12-02
Added test case for a future exactness ranking score

Toflar:


Toflar:
Here we go, this was a lot easier with the new abstraction now ;) Happy for reviews @daun

daun:
Looks great 🎉 Nothing much to add besides that. The new rankings all seem reasonable!

Toflar:
Noticed that `exactness` is basically the same as 0 typos. So while on it I also implemented the `typo` ranker. Also moved the ranker validation to the `Configuration` itself. This looks good now to me 😊

PR #126, opened 2024-12-03
Benchmark ergonomics

daun:
A few tweaks to the bin/benchmark scripts.

- Shortened script names (less typing 🥳)
- Increase memory limit (couldn't get it to run otherwise)
- Clear index before indexing using new method
- Add command line options for
  - Limiting the amount of movies to index `--limit=1000`
  - Re-inserting all existing documents to test update performance: `--update`
  - Changing the search query term `--query="lorem ipsum"`
  - Debugging the search results `--debug`
- Display search runtime in `ms` instead of `s`
- Reordered time vs. memory (I figured time's more relevant most of the time)
- Add newline at the end to avoid php adding a `%` after each run

Screenshot 2024-12-03 at 12 07 13


Toflar (approved):

Toflar:
Thanks, Philipp!

PR #129, opened 2024-12-04
Some more sqlite optimizations

daun:
From a recent project, I learned I've been doing SQLite pragmas all wrong 😵‍💫 The page size change isn't applied when the database is currently in WAL mode. We need to turn that off and then re-enable it after changing the page size.

I'm seeing about a **20% increase in indexing speed** with that (94s instead of 114s). No effect on query speed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,16 +29,16 @@ Note that some implementation details (e.g. libraries used) referenced in this b
 
 ## Performance
 
-Performance depends on many factors but here are some ballpark numbers based on indexing the [~32k movies fixture by 
-MeiliSearch][MeiliSearch_Movies] and the test files in `bin` of this repository:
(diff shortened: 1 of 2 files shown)


Toflar:
Nice! ❤️  Do you want to adjust the stats in the `README.md` as well?

daun:
@Toflar I've updated the readme section on performance. I couldn't help myself and rewrote it a bit for clarity from an "outsider's" perspective, but let me know if you prefer just a raw numbers update and I'll happily do that too.

Toflar:
Great stuff! Thanks! No, I love the fact that you help make it your own! It's open source software. What makes it so powerful is when different people with different backgrounds bring their ideas to the table 😊

PR #130, opened 2024-12-05
Cache sqlite custom function calls

daun:
It appears that SQLite will call a custom function twice when selecting its result as a column and sorting by that column. There's an option of [marking functions as deterministic](https://www.php.net/manual/en/sqlite3.constants.php#constant.sqlite3-deterministic), but the query planner isn't required to honor that "request". In our case, it has no effect on the number of function calls.

What did have an effect (although less than expected) was caching/memoizing the calls on the PHP side.

That brings about a 2-5% increase in query speed depending on complexity and typos.

### Benchmarks

| :--- | :--- | :--- |
| amakin dkywalker | 162.1 ms ±   1.6 ms | 159.3 ms ±   1.7 ms |
| famly dadd | 688.3 ms ±  13.3 ms | 654.7 ms ±   7.2 ms |
| once twice thrice | 270.1 ms ±  10.4 ms | 256.7 ms ±   5.4 ms |
| lighting | 194.9 ms ±   3.0 ms | 190.5 ms ±   1.9 ms |

### DIFF
--- a/src/Internal/Levenshtein.php
+++ b/src/Internal/Levenshtein.php
@@ -8,7 +8,7 @@ use Toflar\StateSetIndex\DamerauLevenshtein;
 
 class Levenshtein
 {
-    public static function damerauLevenshtein(string $string1, string $string2, bool $firstCharTypoCountsDouble): int
+    public static function damerauLevenshtein(string $string1, string $string2, int|bool $firstCharTypoCountsDouble): int
     {
         $distance = DamerauLevenshtein::distance($string1, $string2);
 
@@ -19,7 +19,7 @@ class Levenshtein
         return $distance;
     }
 
-    public static function maxLevenshtein(string $string1, string $string2, int $maxDistance, bool $firstCharTypoCountsDouble): bool
+    public static function maxLevenshtein(string $string1, string $string2, int $maxDistance, int|bool $firstCharTypoCountsDouble): bool
     {
         $distance = DamerauLevenshtein::distance($string1, $string2, $maxDistance + 1);
 
--- a/src/Traits/MemoizationTrait.php
+++ b/src/Traits/MemoizationTrait.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Loupe\Loupe\Traits;
+
+trait MemoizationTrait
+{
+    /**
+     * @var array<mixed>
+     */
+    private array $memo = [];
+
+    private function memoize(callable $compute): mixed
+    {
+        $namespace = serialize($compute);
+
+        return function() use ($namespace, $compute) {
+            $args = func_get_args();
+            $key = md5($namespace . serialize($args));
+            return $this->memo[$key] ??= call_user_func_array($compute, $args);
+        };
+    }
+}
--- a/src/Traits/MemoizationTrait.php
+++ b/src/Traits/MemoizationTrait.php
@@ -7,18 +7,20 @@ namespace Loupe\Loupe\Traits;
 trait MemoizationTrait
 {
     /**
-     * @var array<mixed>
+     * @var array<string, mixed>
      */
     private array $memo = [];
 
-    private function memoize(callable $compute): mixed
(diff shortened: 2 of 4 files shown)


daun:
Something off about the CI tests, need to investigate a bit :)

src/Internal/Levenshtein.php
```diff
@@ -8,7 +8,7 @@
 
 class Levenshtein
 {
-    public static function damerauLevenshtein(string $string1, string $string2, bool $firstCharTypoCountsDouble): int
+    public static function damerauLevenshtein(string $string1, string $string2, int|bool $firstCharTypoCountsDouble): int
```
daun, line 11: These `int|bool` and `int|string` reflect the fact that sqlite does some magic type coercion under the hood, which gets lost in translation during memoization. I actually like this, since it makes it explicit that we're getting sql-type `int` values instead of actual php-type `bool` values, e.g.

src/Traits/MemoizationTrait.php
```diff
@@ -0,0 +1,24 @@
...
+{
+    private function memoize(callable $compute): callable
+    {
+        return function () use ($compute) {
+            /**
+             * @var array<string, mixed>
+             */
+            static $cache = [];
+            $args = \func_get_args();
+            $key = implode('--', $args);
```
daun, line 17: We're getting away with this simple way of "hashing" the arguments since all sqlite function arguments are scalar values, so no need to json-encode or serialize anything :)

daun:
Ready for review

Toflar (commented):
I like the fact of caching the things but we definitely need a different approach. We are already leaking memory with static caches elsewhere and we need to provide a way to prevent that from happening for long-running PHP workers such as frankenphp, swoole, roadrunner etc.

I don't like the trait. I do like traits when used for the right purpose. The one here is not one of them.
What do you think of adding a central static cache utility that we can use everywhere and that is reset upon object destruction? (I imagine maybe we can work with `spl_object_id()` or so) 
If that doesn't work, then we have to actively reset the caches after `Loupe::addDocuments()` and `Loupe::search()`.

daun:
@Toflar Great point about long-running workers. Self-destroying objects sound like a good option for that type of thing. We might even get away with a good old destructor method. The proper destruction and lack of memory leaks will be a challenge to test, I guess.

Toflar:
Yeah I was thinking about the destructor too but we'll have to pass around the spl object id for all those SQLite methods. But I figured it would be better to tackle this now than adding yet another different layer 🙈

Toflar:
Also, I love the fact that we have multiple search benchmarks. I would like to investigate what's up with `famly dadd` that it takes so much longer compared to others. Maybe we should just take your matrix to the bench files right away? What do you think?

daun:
I was toying around with different combinations of long/short/misspelled words and for some reason `famly dadd` was a good indicator of performance. We could check what the tokenizer will stem this to, maybe it ends up super generic. I tried searching for `the` on the 32K movies and it took 2 minutes, I think 🤠

daun:
We can think of a few more/better search terms and create a test matrix, sure! I'm using [hyperfine](https://github.com/sharkdp/hyperfine) locally, maybe there's something similar in PHP land. Or we just add a composer script, but contributors need to set up a local benchmarking utility themselves?

Toflar:
Same, same. I also used `hyperfine` - that's perfect. I meant adding the test cases to the bench files you have updated.
The `the` use case is another problem I don't know how to fix but we have to. That's an easy attack vector 😢

daun:
> The `the` use case is another problem I don't know how to fix but we have to. That's an easy attack vector 😢

Right, haven't thought about it from that angle... Stop words can help a bit here, but they're optional and something like `when` or `with` is similarly well-suited for DoSing and might not make sense on every list. Mmh...

daun:
Marking this as a draft until we come up with a better strategy for caching intermediate results.

daun:
Closed in favor of #136

PR #132, opened 2024-12-07
Highlight stop words if they appear in the query

daun:
Slight adjustment to how the highlighter handles stop words.

Currently, stop words are never highlighted. This change will **highlight stop words if they are part of the query and they occur next to other words**. The goal is to improve the match between what people searched for and what gets highlighted. I found this to be most useful in the movie example, where there is lots of `The` and `An`.

### Examples

| Query | Before | After |
|:---|:---|:---|
| Pirates of the Caribbean: The Curse of the Black Pearl | <mark>Pirates</mark> of the <mark>Caribbean</mark>: The <mark>Curse</mark> of the <mark>Black Pearl</mark> | <mark>Pirates of the Caribbean</mark>: <mark>The Curse of the Black Pearl</mark> |
| a table for two | I booked a <mark>table</mark> for <mark>two</mark>. | I booked <mark>a table for two</mark>. |
| racing to a boxing match | While <mark>racing</mark> to a <mark>boxing match</mark>... | While <mark>racing to a boxing match</mark>... |

### Prior art

Meilisearch also highlights stop words with the following [note in their docs](https://www.meilisearch.com/docs/reference/api/search#attributes-to-highlight):

> Note: `attributesToHighlight` also highlights terms configured as [synonyms](https://www.meilisearch.com/docs/reference/api/settings#synonyms) and [stop words](https://www.meilisearch.com/docs/reference/api/settings#stop-words).

### DIFF
--- a/src/Internal/Search/Highlighter/HighlightResult.php
+++ b/src/Internal/Search/Highlighter/HighlightResult.php
@@ -7,7 +7,7 @@ namespace Loupe\Loupe\Internal\Search\Highlighter;
 class HighlightResult
 {
     /**
-     * @param array<int, array{start: int, length: int}> $matches
+     * @param array<int, array{start: int, length: int, stopword: bool}> $matches
      */
     public function __construct(
         private string $highlightedText,
@@ -21,7 +21,7 @@ class HighlightResult
     }
 
     /**
-     * @return array<int, array{start: int, length: int}>
+     * @return array<int, array{start: int, length: int, stopword: bool}>
      */
     public function getMatches(): array
     {
--- a/src/Internal/Search/Highlighter/Highlighter.php
+++ b/src/Internal/Search/Highlighter/Highlighter.php
@@ -73,7 +73,7 @@ class Highlighter
     }
 
     /**
-     * @param array<array{start:int, length:int}> $matches
+     * @param array<array{start:int, length:int, stopword:bool}> $matches
      * @return array{starts: array<int>, ends: array<int>}
      */
     private function extractSpansFromMatches(array $matches): array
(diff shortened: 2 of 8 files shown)


daun:
@Toflar What are your thoughts on this one? It only becomes an issue when using stop words, but it's quite noticeable on certain queries.

Toflar:
Oh sorry, I missed this totally! Nice work! This makes total sense to me, I just wonder how MeiliSearch handles this. Did you research maybe and want to put the notes here for future reference?

daun:
@Toflar Meilisearch also highlights stopwords. They have the below note in their docs. I don't think it makes sense to highlight all stop words as you'll have the highlights littered with `a` and `the`, but the note is pretty vague and can be read either way. Updated the PR comment with the note and a link to Meilisearch docs as well.

> Note: `attributesToHighlight` also highlights terms configured as [synonyms](https://www.meilisearch.com/docs/reference/api/settings#synonyms) and [stop words](https://www.meilisearch.com/docs/reference/api/settings#stop-words).

Toflar:
Thanks a lot for yet another awesome contribution! Thinking about providing some default stop word list. Something like we have for `TypoTolerance` already, maybe?

* `$configuration->withStopWords(Stopwords::defaultList())` (enabled by default, who does not want to have stop words?)
* `$configuration->withStopWords(Stopwords::disable())`
* `$configuration->withStopWords(Stopwords::withAddedToDefault('foo', 'bar', 'baz'))`
* `$configuration->withStopWords(Stopwords::withRemovedFromDefault('foo', 'bar', 'baz'))`

just thinking out loud.

daun:
@Toflar A default stop word list sounds great for efficiency. Probably as an opt-in setting, though — you can easily get into trouble with multi-language setups. I can imagine somebody indexing french documents and not finding documents about tea (`thé`). You're probably only generating support requests with a default-enabled set of stop words 🙃 But providing the list would make sense, I think.

Toflar:
Which is why I think stop words lists should be language dependent ;)

PR #136, opened 2024-12-10
Implemented a static cache making sure it's reset on object destruction

Toflar:
Alternative implementation to https://github.com/loupe-php/loupe/pull/130.

It's not really beautiful but static caches never are :/ 
@daun would love to get some feedback on the stats you get and also regarding the implementation of course.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -20,4 +20,9 @@
             <directory>tests/Functional</directory>
         </testsuite>
     </testsuites>
+
+    <extensions>
+        <bootstrap class="Loupe\Loupe\Tests\PhpUnit\Extension\LoupeTestsExtension">
+        </bootstrap>
+    </extensions>
 </phpunit>
--- a/src/Internal/Levenshtein.php
+++ b/src/Internal/Levenshtein.php
@@ -19,7 +19,7 @@ class Levenshtein
         return $distance;
     }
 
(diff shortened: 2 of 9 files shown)


daun:
Looks great! It's not that pretty, but if it's faster, who cares 🤠 Will run a few benchmarks in a minute.

One question: what's the reason the call to `StaticCache::enterContext($this)` must happen inside each method call to the Loupe instance?  As opposed to doing it once in the constructor.

Toflar:
> One question: what's the reason the call to `StaticCache::enterContext($this)` must happen inside each method call to the Loupe instance? As opposed to doing it once in the constructor.

Think of an application that works with two or more `$loupe` instances at the same time in the same process:

```
$loupe1->addDocument(...);
$loupe2->search(...);
$loupe3->addDocuments(...);
...
```

daun:
I see, that would allow switching back and forth without clearing the cache in between. Nice.

daun:
I'm seeing about a 1-2% speed bump for queries across the board. The very obvious question is of course if that's worth all the acrobatics and maintenance burden going forward.

What are you getting?

Toflar:
Similar 😊  I would argue that the work is done now anyway and I'm pretty sure there might be other use cases for a cache where we'd again needed such a static cache management. So I'd merge that for now but we can of course still remove it again if we have weird side-effects or encounter any other issues.

daun:
Alright, I'll close the other one.

Toflar:
Thanks for the valuable feedback @daun!

Issue #134, opened 2024-12-09
Prevent "endless" search requests


Toflar:
Apparently, when having lots of documents and searching for a very common word like `the`, you can cause Loupe to crash because it finds too many matching documents and tries to calculate way too much.
We should try to fix that for sure. I don't know what is the best way but probably just an internal absolute limit.

daun:
Just brainstorming some ideas:

- Reject queries that only contain very common words (the, an, a, with) — maybe reuse stop words for that
- Set a hard timeout for query execution time (e.g. 1.000ms)
- Set a hard upper limit on number of processed results (e.g. 10.000)
- Rate limiting (this one's better implemented on the consumer side)
- Cache results for around 1 minute (might also be better to do on the consumer side)

Toflar:
See https://github.com/loupe-php/loupe/pull/137

PR #137, opened 2024-12-10
Prevent "endless" queries

Toflar:
Fixes https://github.com/loupe-php/loupe/issues/134.

I have accidentally pushed this already in https://github.com/loupe-php/loupe/pull/136/files#diff-60e811f23f79f3d9ec47e6aa6599834bd758f7063de8dd1be2b9ae821d2615ea but here's the finalized version.

Tested using `php bin/bench/search.php -q="this is taking so long"`.
Previously: 31283.11
Now: 1235.46

Search results are still "relevant" but again. What does "relevant" even mean with such a query 😊

### DIFF
--- a/src/Internal/Search/Searcher.php
+++ b/src/Internal/Search/Searcher.php
@@ -26,13 +26,6 @@ use Loupe\Loupe\SearchResult;
 
 class Searcher
 {
-    /**
-     * If searching for a query that is super broad like "this is taking so long", way too many
-     * documents are going to match so we have to internally limit those matches to prevent
-     * "endless" search queries.
-     */
-    private const MAX_DOCUMENT_MATCHES = 1000;
-
     public const CTE_TERM_DOCUMENT_MATCHES_PREFIX = '_cte_term_document_matches_';
 
     public const CTE_TERM_MATCHES_PREFIX = '_cte_term_matches_';
@@ -40,6 +33,13 @@ class Searcher
     public const DISTANCE_ALIAS = '_distance';
 
     public const RELEVANCE_ALIAS = '_relevance';
+
+    /**
+     * If searching for a query that is super broad like "this is taking so long", way too many
+     * documents are going to match so we have to internally limit those matches to prevent
+     * "endless" search queries.
+     */
+    private const MAX_DOCUMENT_MATCHES = 1000;
 
     /**
      * @var array<string, array{cols: array<string>, sql: string}>


daun:
Does this limit the total number of considered results to 1000? Or is that per term?

How does this behave if I have a total of 1.500 documents in the index and all matches are kind of bad (stemmed, low-ranking attribute, etc.) but the perfect match is the 1200th document with a perfect title match?

Toflar:
Would need to be tested but I guess you just won't find that document then.

daun:
If that limitation is per term, I think that's a reasonable compromise to make.

If the limit is for the whole result set, that might lead to some rather suboptimal search results in the long term, given that multiple terms are conceptually joined by OR. If a query starts with multiple generic terms and ends with a very specific term (The Long Day of Rübenthaler), it would probably fetch 1000 documents for the generic terms first, and then ignore the one result for the specific term since it's over the limit already. Not sure I correctly understood the logic here, though 🤠

Toflar:
Unfortunately, I discovered even more issues. The `LIMIT` query is currently applied completely incorrectly. I tried fixing that for half a day but I gave up for now.

Toflar:
And even creating a test case for this issue here does not work. Bad day at work :/

Toflar:
Found out that it's not a general bug with the LIMIT query but this PR in its current state indeed causes issues. I will add a test case.

Toflar:
However, I don't really know if there's a different way to fix this issue. The query matches too many documents = you cannot expect it to find the most relevant document because in order to do that, it would have to search **all** of them which is exactly what you want to prevent. I think the same happens in MeiliSearch: https://www.meilisearch.com/docs/learn/resources/known_limitations#maximum-number-of-results-per-search

Toflar:
Again, you cannot prevent that from happening. What is relevant and what's not is determined on the **entire** result set. If you strip parts of it, then it's always possible that the most relevant gets stripped. You can try to improve the probability on some wild guesses like "longer terms are probably more distinct" etc. or we could again count the term frequency and order somehow by that but that's just a whole different story. I think we should first have a security guard which prevents Loupe from easily getting crashed which is what this PR does. Then we can still try to come up with ways to optimize this "security guard".

Toflar:
See https://github.com/loupe-php/loupe/pull/140 for a test case if you want to give it a try somehow.

daun:
Sounds like a good strategy! Lock things down to avoid the security issue and optimize later. I'll see if I find a minute to look into the test draft :) But the optimization shouldn't be a blocker here.

PR #141, opened 2024-12-17
Reworked relevance rankers and fixed proximity ranker

Toflar:
Reworked the way rankers get their information, it's now all in nice objects making it easier to understand what's going on.

Moreover, I found that the `Proximity` ranker was buggy because it did not consider the attribute. It's best seen in the `SearchTest::testRelevanceAndRankingScoreWithAttributeWeights()`.

Previously, when searching for `game of life` the following document got a relevance score of `0.8034`:
```
[
     'id' => 3,
     'title' => 'Learning to game',
     'content' => 'What life taught me about learning',
]
```

This is nonsense but happened because `game` occurs at position 3 within `title` and `life` at position 2 within `content`. Without distinguishing the different attributes, this results in close proximity.

With this PR, this now gets downgraded to `0.64323`. 

Would love to have a review @daun - maybe you see false-positives in the adjusted tests. Also, the information is now not passed on using references anymore to the rankers. As I said, it should have no negative effect because objects are also references but you might want to run your bench again. For me, this PR is actually quite a bit faster because information can be re-used 😉 I am now at about 105ms instead of 120ms as stated in the `README`. Would love to learn about your results.

TODO:

- [ ] Update `README.md` because Loupe got faster with this PR?

### DIFF
--- a/src/Internal/Search/Ranking/RankingInfo.php
+++ b/src/Internal/Search/Ranking/RankingInfo.php
@@ -1,0 +1,63 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Loupe\Loupe\Internal\Search\Ranking;
+
+final class RankingInfo
+{
+    /**
+     * @var array<string>
+     * @var array<string>
+     */
+    private function __construct(
+        private array $rankingRules,
+        private array $searchableAttributes,
+        private TermPositions $termPositions
+    ) {
+    }
+
+    /**
+     * Example: A string with "3:title,8:title,10:title;0;4:summary" would read as follows:
+     * - The query consisted of 3 tokens (terms).
+     * - The first term matched. At positions 3, 8 and 10 in the `title` attribute.
+     * - The second term did not match (position 0).
+     * - The third term matched. At position 4 in the `summary` attribute.
+     *
+     * @param string $termPositions A string of ";" separated per term and "," separated for all the term positions within a document
+     */
+    public static function fromQueryFunction(string $searchableAttributes, string $rankingRules, string $queryTokens, string $termPositions): self
+    {
+        $info = new self(explode(':', $rankingRules), explode(':', $searchableAttributes), TermPositions::fromQueryFunction($termPositions));
+
+        // TODO: Who needs those?
+        $queryTokens = explode(':', $queryTokens);
+
+        return $info;
+    }
+
+    /**
+     * Returns the ranking rules in order as defined in Configuration.
+     *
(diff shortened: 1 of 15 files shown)


src/Internal/Search/Ranking/RankingInfo.php
```diff
@@ -0,0 +1,63 @@
...
+     * - The third term matched. At position 4 in the `summary` attribute.
+     *
+     * @param string $termPositions A string of ";" separated per term and "," separated for all the term positions within a document
+     */
+    public static function fromQueryFunction(string $searchableAttributes, string $rankingRules, string $queryTokens, string $termPositions): self
+    {
+        $info = new self(explode(':', $rankingRules), explode(':', $searchableAttributes), TermPositions::fromQueryFunction($termPositions));
+
+        // TODO: Who needs those?
+        $queryTokens = explode(':', $queryTokens);
```
daun, line 34: I think the reason this is here is to prepare for an exactness ranking so it can check exact matches vs. stemmed matches. But it might be covered elsewhere in this PR?

tests/PhpUnit/Subscriber/AssertStaticCacheIsEmptySubscriber.php
```diff
@@ -13,6 +13,6 @@ class AssertStaticCacheIsEmptySubscriber implements FinishedSubscriber
 {
     public function notify(Finished $event): void
     {
-        Assert::assertTrue(StaticCache::isEmpty(), 'Static cache must be empty in: ' . $event->test()->id());
+        //Assert::assertTrue(StaticCache::isEmpty(), 'Static cache must be empty in: ' . $event->test()->id());
```
daun, line 16: I've been having issues with this in testing as well. Only way I found was by disabling it, too 🙃

daun:
Benchmark-wise, I'm seeing no difference (139.1 vs 139.4). Which is amazing, considering the improved maintainability of the ranking algorithm.

src/Internal/Search/Ranking/RankingInfo.php
Toflar: They were present before, we'll leave them in for now.

tests/PhpUnit/Subscriber/AssertStaticCacheIsEmptySubscriber.php
Toflar: Only when they fail - I need to check that again but for now I have to re-enable.

PR #144, opened 2025-01-16
Use binary format for state set cache for better performance and OPcache exclusion

Toflar:
~~I want to disable storing the state set in OPcache by default for the reasons described in the docs of this PR.~~

~~However, the current solution using `ini_set('opcache.enable', '0');` won't work because this affects all other scripts loaded after the `state_set.php` as well. However, I couldn't find a way to disable OPcache for just one specific file while still dumping that file as pure PHP because that's still a lot faster than loading all the states from SQLite...~~

~~Any hints welcome 😊~~

Using `pack()` and `unpack()` was way faster in my tests than just a regular PHP array so let's drop OPcache support completely for better interoperability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Performance depends on many factors but here are some ballpark numbers based on 
 [~32k movies fixture][MeiliSearch_Movies] provided by MeiliSearch.
 
 * **Indexing** will take a little over **90 seconds** (~350 documents per second)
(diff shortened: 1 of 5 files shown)


ausi:
Did you try storing it in binary format using something like

```php
$states = [1 => true, 7 => true, 123 => true];
file_put_contents(…, bin2hex(pack('N*', ...array_keys($states))));
```

And then loading it back in using unpack:

```php
$data = unpack('N*', file_get_contents(…));
$states = array_combine($data, array_fill(0, \count($data), true));
```

PR #146, opened 2025-01-23
State Set using Bloom Filter

ausi:
A quick POC. Looks like the query is a tiny bit slower now. Needs to be investigated though.

### DIFF
--- a/src/Internal/StateSetIndex/StateSetBloomFilter.php
+++ b/src/Internal/StateSetIndex/StateSetBloomFilter.php
@@ -1,0 +1,159 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Loupe\Loupe\Internal\StateSetIndex;
+
+use Toflar\StateSetIndex\StateSet\StateSetInterface;
+
+class StateSetBloomFilter implements StateSetInterface
+{
+    private string $data;
+
+    private float $falsePositiveProbability;
+
+    private int $hashCount;
+
+    /**
+     * @var array<int,true>|null
+     */
+    private array|null $states;
+
+    private function __construct()
+    {
+    }
+
+    public function add(int $state): void
+    {
+        if ($this->states !== null) {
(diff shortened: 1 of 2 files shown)


Toflar:
I have also implemented a bloom filter myself but I never managed to get it faster than the simple memory lookup :)

src/Internal/StateSetIndex/StateSetBloomFilter.php
```diff
@@ -0,0 +1,159 @@
...
+        return ($this->data[$byteOffset] & \chr(1 << ($offset % 8))) !== "\x00";
+    }
+
+    /**
+     * @return \Generator<int>
+     */
+    private function hashIterator(int $state): \Generator
+    {
+        $combinedHash = ($state << 4) % 0xFFFFFFFF;
+        $secondHash = crc32(pack('N', $combinedHash));
```
Toflar, line 146: Could use xxh3?

ausi:
> I never managed to get it faster than the simple memory lookup :)

Yeah, not sure if it is even possible. But I’d also like to try a larger index size for the performance tests as the current test case has only around 50k states. Does loupe have a maximum capacity it was designed for? The *movies.json* has 16MB. How about 100MB or 1GB?

Toflar:
Haven't tried. Would need to find a test set :) The movies are from Meilisearch.

Toflar:
I think you could test with the https://musicbrainz.org/doc/MusicBrainz_Database dataset. It contains over 34 million recordings (songs).
I'm pretty sure that at some point, the bloom filter will become more efficient than the in-memory logic. That being said, I would think that it might be better to incorporate the bloom filter logic into the current state set and not have a `StateSetBloomFilter` that implements the interface. It's only relevant for the `has()` method anyway. There could be an internal threshold where the bloom filter kicks in vs. the in-memory stateset. E.g. if `count($this->states) > ...` then store and lookup differently or so.

PR #147, opened 2025-01-29
Add new test at testStopWordSearch

zonky2:
If we use default sorting (relevance) we have follow error

`Unable to prepare statement: no such table: _cte_term_document_matches_0`


Toflar:
Fixed in https://github.com/loupe-php/loupe/pull/148.
PS: You are not supposed to use the `Searcher::...` class from the outside. This is internal. Anything in the `Internal` namespace might change in any version without being considered a BC break. Using `_relevance` as a string is the API as documented :)

PR #159, opened 2025-03-19
Fixed more filter bugs and introduced a BETWEEN operator

Toflar:
Turns out there were more bugs with the latest implementation and the `multi-attribute >= 5 AND multi-attribute <= 8` just didn't work out to be practical.
So I had to fix this and introduce a `BETWEEN` filter to achieve this logic instead.


alexander-schranz (approved):

alexander-schranz:
Can confirm this solve both bug we currently had since Loupe 0.9.5 with the SEAL test suite.

PR #163, opened 2025-03-26
Ensure correct relevance ranking for all returned documents

### LINKED ISSUE #142: Optimize ranking of documents past the query limit
The max result limit introduced in #137 can be optimized to allow ranking documents that are partly past the limit on some terms.

The limit of `1000` is currently applied to each term CTE, i.e. we aren't skipping search terms even if all previous terms combined would already match 1000 documents.

The issue however is that the compiled match positions are off if a document is already past the limit on some terms but inside the limit on others. Hence, we don't have a chance of compiling accurate ranking scores on those documents, even if they are included in the result.

### Example

Indexed documents

- 2000 documents with content 'dog'
- 1 document with content 'dog house'

Search query

- `dog house`

Compiled match positions

- `1:content;0` for the first 2000 documents
  - Makes sense, as the second term `house` isn't found in the content
- `0;2:content` for the last 1 document
  - Doesn't make sense, as both terms are part of the content
  - Explained by the fact that the document was past the limit on the first term, hence `0`
  - This prevents the document from properly ranking

Goal

- Make sure the `dog house` document ranks highest for the `dog house` query, even if partly past the limit

daun:
- Make sure we get accurate ranking data for all documents, including those past the 1000-documents-per-term limit
- Matching 1.000 documents for term A and 100 documents for term B will correctly calculate relevance for all 1.100 docs
- The idea is to query the document-terms matches for all documents after combining them across terms
- Decreases performance by around 40ms on most queries, which isn't amazing but preferable to incorrectly calculated ranking scores. Maybe the performance can still be improved, but I guess correctness trumps speed...
- Closes #142

### DIFF
--- a/src/Internal/Search/Searcher.php
+++ b/src/Internal/Search/Searcher.php
@@ -26,13 +26,6 @@ use Loupe\Loupe\SearchResult;
 
 class Searcher
 {
-    /**
-     * If searching for a query that is super broad like "this is taking so long", way too many
-     * documents are going to match so we have to internally limit those matches to prevent
-     * "endless" search queries.
-     */
-    private const MAX_DOCUMENT_MATCHES = 1000;
-
     public const CTE_TERM_DOCUMENT_MATCHES_PREFIX = '_cte_term_document_matches_';
 
     public const CTE_TERM_MATCHES_PREFIX = '_cte_term_matches_';
--- a/src/Internal/Search/Searcher.php
+++ b/src/Internal/Search/Searcher.php
@@ -7,17 +7,11 @@ namespace Loupe\Loupe\Internal\Search;
 use Doctrine\DBAL\ArrayParameterType;
 use Doctrine\DBAL\Query\QueryBuilder;
 use Doctrine\DBAL\Result;
-use Location\Bounds;
 use Loupe\Loupe\Internal\Engine;
-use Loupe\Loupe\Internal\Filter\Ast\Concatenator;
-use Loupe\Loupe\Internal\Filter\Ast\Filter;
-use Loupe\Loupe\Internal\Filter\Ast\GeoBoundingBox;
-use Loupe\Loupe\Internal\Filter\Ast\GeoDistance;
-use Loupe\Loupe\Internal\Filter\Ast\Group;
-use Loupe\Loupe\Internal\Filter\Ast\Node;
+use Loupe\Loupe\Internal\Filter\Ast\Ast;
 use Loupe\Loupe\Internal\Filter\Parser;
 use Loupe\Loupe\Internal\Index\IndexInfo;
-use Loupe\Loupe\Internal\LoupeTypes;
+use Loupe\Loupe\Internal\Search\FilterBuilder\FilterBuilder;
 use Loupe\Loupe\Internal\Tokenizer\Token;
 use Loupe\Loupe\Internal\Tokenizer\TokenCollection;
 use Loupe\Loupe\Internal\Util;
@@ -46,10 +40,14 @@ class Searcher
      */
     private array $CTEs = [];
 
+    private Ast $filterAst;
+
     /**
      * @var array<string, bool>
      */
     private array $geoDistanceSelectsAdded = [];
+
+    private bool $multiAttributeJoinAdded = false;
 
     private QueryBuilder $queryBuilder;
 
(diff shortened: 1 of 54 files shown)


daun:
Feel free to rebase the whole thing or cherrypick from the messy commit history :)

src/Internal/Search/Searcher.php
```diff
@@ -278,16 +278,31 @@ private function addTermDocumentMatchesCTE(Token $token, ?Token $previousPhraseT
 
         $cteSelectQb->from(IndexInfo::TABLE_NAME_TERMS_DOCUMENTS, $termsDocumentsAlias);
 
+        // Get documents that match any of our terms
+        $documentConditions = [];
+        foreach ($this->getTokens()->all() as $t) {
+            $cteName = '_cte_term_documents_' . $t->getId();
```
Toflar, line 284: Can we externalize this logic into a method which we then re-use in `addTermDocumentsCTE()`?

Toflar (commented):
Hey Philipp, thanks so much for working on this! ❤️  Looks great!
Did you test the performance impact using the `bin/bench` files?

FYI: I'm on vacation soon, might take a while to reply, I hope that's okay 😊

daun:
Yes, I've used the bench files in `bin`. I've updated the PR description with before/after measurements. Looks like the hit is around 20-25ms per search term, or 20-40% in relative terms. Not great, but we can probably fumble ourselves back towards better performance :)

| :--- | :--- | :--- |
| amakin dkywalker | 139.5 ms ± 7.6 ms | 180.3 ms ±     ms |
| famly dadd | 203.4 ms ± 7.7 ms | 246.7 ms ± 8.4 ms |
| once twice thrice | 179.3 ms ± 3.4 ms | 278.2 ms ± 5.8 ms |
| lighting | 160.6 ms ± 4.6 ms | 223.1 ms ± 6.1 ms |

daun:
Nice, enjoy the vacation! We can pick things up when you're back, no rush here.

Toflar:
Yeah, let's also see how this will play with https://github.com/loupe-php/loupe/pull/166 😊 
I'm okay with a little performance penalty if it turns out to provide better results.

Toflar:
I have merged https://github.com/loupe-php/loupe/pull/166 into develop now. I wanted to do this with priority because it is fixing bugs (that are covered in tests in the mentioned PR now). Your PR needs a rebase because of my `Cte` class refactoring but the query filter logic has remained largely untouched so it should be fairly straightforward, I hope. Wanna give it a try?

daun:
Nice! Happy to get it rebased but I'm hopelessly swamped with client deadlines this week. If you'd rather get it over with before next week, feel free to give it a go yourself!

Toflar:
No worries, it's not urgent. I can require `dev-develop` 😊  I would love to have you look at it just so that at least I have a second pair of eyes checking what I did 😊

daun:
@Toflar Rebased onto develop. Ready for review :) The new changes on develop look really good to me. They especially seem to fix an issue with number/string sorting of the user id fields which makes the order more predictable I think.

src/Internal/Search/Searcher.php
```diff
@@ -346,6 +362,50 @@ private function addTermDocumentMatchesCTEs(TokenCollection $tokenCollection): v
...
+        $termMatchesCTE = $this->getCTENameForToken(self::CTE_TERM_MATCHES_PREFIX, $token);
+
+        if (!$this->hasCTE($termMatchesCTE)) {
+            return;
+        }
+
+        $termsDocumentsAlias = $this->engine->getIndexInfo()->getAliasForTable(IndexInfo::TABLE_NAME_TERMS_DOCUMENTS);
+
+        $cteSelectQb = $this->engine->getConnection()->createQueryBuilder();
+        $cteSelectQb->addSelect('DISTINCT ' . $termsDocumentsAlias . '.document');
```
Toflar, line 377:
```suggestion
        $cteSelectQb->select($termsDocumentsAlias . '.document')->distinct()
```

is what I have used elsewhere :)

Toflar (changes requested):
Looks good to me! Just a minor nitpick, then I'll merge :)

src/Internal/Search/Searcher.php
daun: Makes sense!

Toflar:
Thanks a lot @daun! ❤️

PR #164, opened 2025-03-26
Make document limit configurable

daun:
- Allow customizing the 1000-documents-per-term limit
- Needs to be rebased if/when merging #163 but wanted to get your opinion on making this configurable
- Meilisearch has a similar setting [`maxTotalHits`](https://www.meilisearch.com/docs/reference/api/settings#update-pagination-settings) which applies to the whole result set, so is a bit different in scope
  - If we named this `maxTotalHits` as well and also applied it to the final result query as well, we'd have the same concept. WDYT? Might be good to keep this in sync conceptually.

### DIFF
--- a/src/Internal/Search/Searcher.php
+++ b/src/Internal/Search/Searcher.php
@@ -29,13 +29,6 @@ class Searcher
     public const RELEVANCE_ALIAS = '_relevance';
 
     /**
-     * If searching for a query that is super broad like "this is taking so long", way too many
-     * documents are going to match so we have to internally limit those matches to prevent
-     * "endless" search queries.
-     */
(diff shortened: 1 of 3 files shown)


Toflar:
Sure, we should definitely do that! 👍

daun:
@Toflar Would you rather introduce this as-is — with a new `getMaxHitsPerTerm` config — or should we stick with Meilisearch and introduce a `maxTotalHits` config? I'd suggest the `maxTotalHits` solution as it's conceptually easier to understand and doesn't require knowledge of the internals.

Toflar:
I agree we should name it `maxTotalHits` and also apply it to the total count. This is easier to understand from a user's perspective 👍

daun:
Closing in favor of #172

PR #165, opened 2025-03-26
Crop results and highlights

### LINKED ISSUE #143: Search snippets / cropped highlights
Just wondering if you'd be willing to support a feature for search snippets — i.e. highlighted search results that crop the returned text to areas around matched terms.

> <mark>Lorem ipsum</mark> dolor sit amet, consetetur [...] no sea takimata sanctus est <mark>lorem</mark> est <mark>ipsum</mark> dolor sit amet. <mark>Lorem ipsum</mark> dolor sit amet, consetetur [...] dolore te feugait nulla facilisi <mark>lorem ipsum</mark> dolor sit amet, consectetuer [...]

I've [built this feature](https://github.com/daun/statamic-loupe/blob/main/src/Search/Snippets.php) for an integration with a third-party cms a while ago — it was easy enough because I could just pick up where Loupe's internal highlighter left things. But it got me thinking if this would be interesting to have in the engine itself.

For what it's worth, Meilisearch calls this [attributes to crop](https://www.meilisearch.com/docs/reference/api/search#attributes-to-crop).

Suggested Loupe API:

```php
$searchParams->withAttributesToCrop(['title' => 10, 'summary' => 6])
```

daun:
- [x] Create Formatter class to take care of match positions, highlighting and cropping
- [x] Move highlighting logic into Formatter class
- [x] Move match position logic into Formatter class
- [x] Implement cropping logic
- [x] Add tests for cropping logic
- [x] Move highlighting and cropping tests into their own classes
- [x] Make crop length configurable by field
- Closes #143

### Notes

- Terminology and concept matches Meilisearches [attributes to crop](https://www.meilisearch.com/docs/reference/api/search#attributes-to-crop)
- The single deviation is that Loupe measures crop length in characters, while Meilisearch measures in words
- I find characters to be a bit more intuitive
- Cropping when not also highlighting uses a rather simple trick: highlight matches, crop around the matches, and then remove the highlight markers again. Conceptually much easier to reason about than a custom cropping implementation independent of highlighting
- I didn't use the context-cropper composer package to avoid introducing a new dependency on `symfony/string`

### Usage

```php
    ->withAttributesToCrop(['title', 'summary'], 30, '[...]');
```

With custom crop length and marker:

```php
    ->withAttributesToCrop(['title', 'summary'], 30, '[...]');
```

With custom crop length for each attribute:

```php
    ->withAttributesToCrop(['title' => 20, 'summary' => 40]);
```


daun:
@Toflar Ready for review :)

Toflar:
Wow, this looks amazing! Really looking forward to having this feature in Loupe! 🥳 
Do I understand it correctly that you don't want to use https://github.com/loupe-php/context-cropper just because of `symfony/string`? Because here are my 2 cents on that: I am absolutely with you on not having an extra dependency, if it is not needed. However, if we end up repeating a lot of things that have already been done and been battle-tested in a library as wide-spread as this, we should take it.

Anyway, if you manage without `symfony/string` even better. But I would still like to have the formatting logic in its own https://github.com/loupe-php/context-cropper package, rather than Loupe. Why? Because I also need the context-cropping logic in another project that does not have anything to do with Loupe at all 😄 And I don't want to include Loupe there (with all its dependencies) just for the highlighting utilities. Context highlighting - and also cropping - sounds like it can be super useful in completely different applications.

We could outsource the entire formatting (highlighter, matcher, cropper) into its own library. Actually I don't think "context-cropper" is such a bad name for that but I would also be okay with dropping that in favor of another library name? What do you think?

daun:
If I'm being honest, my main reason to avoid the symfony package is performance paranoia. All those objects being created and cloned can't be amazing when doing this over thousands of results. But I must admit I didn't measure 🤡

Outsourcing the whole `Formatting` folder sounds good. We're currently using the internal `TokenCollection` for this quite a bit, which would need to be extracted somehow as well, or replaced with another thing. Is the tokenizer too far embedded, or could this be a package as well?

Meilisearch calls this whole module (formatting, highlighting, matching) the Matcher, which I find very intuitive. It deals with matches :) So maybe `loupe/matcher`? Not great for discoverability on packagist, but in this context rather descriptive I think. Up to you!

There was one edge case with overlapping context that currently isn't handled by the context-cropper package but we can easily backport that into the package.

Toflar:
> If I'm being honest, my main reason to avoid the symfony package is performance paranoia. All those objects being created and cloned can't be amazing when doing this over thousands of results. But I must admit I didn't measure 🤡

You should then 😉 You'll notice that they use less memory than arrays etc. - and the amount here is really not of a concern at all. So the number of objects is very, very likely not a concern. Not to me anyway. But whether or not to introduce a dependency is a different story for sure 😊 



> Meilisearch calls this whole module (formatting, highlighting, matching) the Matcher, which I find very intuitive. It deals with matches :) So maybe `loupe/matcher`? Not great for discoverability on packagist, but in this context rather descriptive I think. Up to you!

I would be totally fine with that! I would also add you as a maintainer there, because like half of the code will be yours anyway 😄 In case you want this, of course!



> There was one edge case with overlapping context that currently isn't handled by the context-cropper package but we can easily backport that into the package.

👍

daun:
Sounds fantastic. If you want, you can create a `matcher` repo and invite me and we'll see where it goes :) Makes sense to start from the current context-cropper implementation and slightly rearrange things to fit a matcher and highlighter in as well.

Toflar:
Done :)

daun:
@Toflar Just took a quick look at this. To properly pull out the Matcher and Highlighter, we would need to also pull out the Tokenizer. Is that something you would consider, or do you prefer that part to be close to the core? From what I gather, its only dependency is on LanguageDetector, and that's really only used in the Tokenizer anyway.

Toflar:
Can you elaborate a bit more on why this needs to be extracted?

daun:
Yes! We're using the Tokenizer to properly identify matching search terms using stemming and language detection. Those resulting Token and TokenCollection objects are then used for passing around information about matched terms inside the Matcher module and in between the Matcher and Highlighter modules. For example, the highlighter needs to know if a matched term is a stopword or not. To decouple things, we would need to tokenize the text inside of Loop, convert the Token objects to raw arrays and then work with raw arrays inside the Matcher module. But maybe I'm missing something that would make this easier?

Toflar:
I see. Makes sense. I think we can easily move the `TokenCollection` in the matcher package. Language detection and stemming would not be relevant for the matcher library so I would go with a `TokenizerInterface` and `TokenInterface` for the matcher library instead. There, we could have very simple implementations for unit tests and standalone use. In Loupe, we'd have our more advanced version with language detection and stemming. Potentially re-using some traits or `AbstractTokenizer` / `AbstractToken` from the matcher library for shared logic.
This would even allow to make the `Tokenizer` configurable in Loupe one day, should the need appear as we only need to allow passing a `TokenizerInterface` instance to the `Configuration` or so 😊  Does that make sense?

daun:
Sounds reasonable, but I'd prefer to leave that type of high-level organization to you in this case. I think I'd be a bit unsure as to what exactly I'm trying to do here. Also, interfaces across modules really confuse me 😅

It might be easiest — in terms of organizing this — to merge this PR as a feature first, and then have you extract the matcher module from there before it goes into master.

The more I look at this, the more complicated it gets. The Matcher module is quite tightly coupled with the Engine and Configuration objects. For example, to determine a match, the matcher must know the `minTokenLengthForPrefixSearch` and `firstCharTypoCountsDouble` config values. Okay, no problem there, we can pass those into the constructor. However, it then uses the TypoTolerance object to calculate the Levenshtein distance for each term. For that, it must reach deep into the machine by going `$this->engine->getConfiguration()->getTypoTolerance()->getLevenshteinDistanceForTerm`. Also, the Levenshtein utility must now be duplicated across modules. It's probably a good idea to decouple this anyway, but I'm a bit overwhelmed here 🙃

Toflar (commented):
I see! Yeah, no worries, then we'll merge this PR as is first and I'll try to extract the logic later 😊

Toflar (changes requested):
There's an organizational issue still: The new tests are in the `Unit` namespace although they are functional (which is clearly indicated by all of them requiring the `FunctionalTestTrait` 😊 So this should be moved. The unit tests should only work with terms directly and compare results, without the entire `Loupe` instance. These are the tests that would then get extracted in the matcher package 😊

daun:
@Toflar Right, I moved things into the correct namespace.

Toflar (approved):

Toflar:
Thank you @daun ❤️

PR #167, opened 2025-04-07
Make configuration serializable

Toflar:
This should help for tools like SEAL that may want to provide some `adapter: loupe://<serialized_config>` configuration value or so 😊 

/cc @alexander-schranz

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -218,5 +218,17 @@ $configuration = \Loupe\Loupe\Configuration::create()
 print_r($logger->getRecords());
 ```
 
+## Serializing and reconstructing configuration
+
+If you need to serialize the configuration of Loupe e.g. to allow configuration using a DSN, you can use the `toString()` and `fromString()` helpers accordingly:
+
+```php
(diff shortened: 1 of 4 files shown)


alexander-schranz:
I have on my map to allow `options` array to be configurable like already done for fields to support "search engine" specific things. That would I think definitly allow this via:

```diff
-        return Configuration::create()
+        return Configuration::fromArray($index->options)
            ->withPrimaryKey($index->getIdentifierField()->name)
            ->withSearchableAttributes(\array_map(fn (string $field) => $this->formatField($field), $index->searchableFields))
            ->withFilterableAttributes(\array_map(fn (string $field) => $this->formatField($field), $index->filterableFields))
            ->withSortableAttributes(\array_map(fn (string $field) => $this->formatField($field), $index->sortableFields));
```

https://github.com/PHP-CMSIG/search/blob/d250362bf6a6b2f84da69a942f7ba136952cda09/packages/seal-loupe-adapter/src/LoupeHelper.php#L142C3-L142C4

As I think this options you menetion can be index specific?

PR #172, opened 2025-04-14
Configurable max total hits

daun:
- Introduce a `maxTotalHits` configuration that limits the number of results across all pages
- Incidentally, this is also applied as limit of results per term for calculating the relevance score
- The total count will also never return more than this number
- Replaces #164 

This is the text used for the docs:

---

### Total result limit

Loupe enforces a maximum number of 1.000 results across all pages to safeguard your index against malicious scraping.
Beyond this limit, no results will be calculated or returned, independent of pagination settings. The limit can be
configured if your application requires a different value.

```php
$config = \Loupe\Loupe\Configuration::create()
    ->withMaxTotalHits(100);
```

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -203,6 +203,17 @@ $typoTolerance = \Loupe\Loupe\Config\TypoTolerance::create()
 ;
 ```
 
+## Total result limit
+
+Loupe enforces a maximum number of 1.000 results across all pages to safeguard your index against malicious scraping.
+Beyond this limit, no results will be calculated or returned, independent of pagination settings. The limit can be
+configured if your application requires a different value.
+
+```php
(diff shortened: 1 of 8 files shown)


Toflar (commented):
I think conceptually, I would place this on the `Configuration` object rather than the `SearchParamters`. Search parameters are dynamic, configuration is static. I know that at the moment it is dynamic anyway but putting it on the `Configuration` would give us the ability to optimize something on the indexing side some day. I cannot think of any at the moment but I would still like to place it where it belongs. A maximum setting doesn't sound like a dynamic thing.
Wdyt?

daun:
Right! Wasn't thinking much about where to put it, but Configuration is definitely the better place :)

Toflar:
Awesome, yet another PR done ✅ Thank you! ❤️

PR #175, opened 2025-04-23
Outsourcing Loupe's formatting etc. to loupe/matcher

Toflar:
This should be the foundational work, only one test is failing which I haven't quite understood.

There's also cleanup work still required in `loupe/matcher` where I started to write something down here: https://github.com/loupe-php/matcher/issues/1

Contributions that help clean this up very welcome!

/cc @daun


daun:
I think the failing test means the custom crop length isn't applied. It should be 20 characters according to the data provider, but it's still the default 50 or so.

Toflar:
Yes but this is exactly what's confusing me. I think your original tests are already wrong? I don't see where the custom crop length is passed in this test? Maybe the naming is not correct?

daun:
I think it‘s a number in the data provider‘s arguments. The results of this test and the previous one should be identical except for the crop length. But I‘m not at the computer at the moment, so I can‘t pull out the exact line of code.

Toflar:
I guess I will merge this now but keep the `dev-main` requirement on the `develop` branch for the time being. Maybe we notice some other issue before the release.

PR #176, opened 2025-04-27
Update wamania/php-stemmer to 4.0

panda-madness:
The 3.0.x versions still require voku/portable-utf8, which causes many deprecation warnings on PHP 8.4.


Toflar:
Thanks!

Issue #179, opened 2025-05-03
Issue with slow sqlite connection optimizations


arnoson:
Creating a new search with `LoupeFactory()::create()` (not the search itself) took a lot of time so I was measuring and realized that [optimizeSQLiteConnection()](https://github.com/loupe-php/loupe/blob/f59b1bc1f3caea29510d20c812fb9dd47105b325/src/LoupeFactory.php#L124) is the cause in my case.

On my local machine running PHP 8.3 they take about 40ms:

| Optimization | Time (ms) |
| --- | --- |
| PRAGMA auto_vacuum = incremental; VACUUM; | 15 |
| PRAGMA journal_mode=DELETE; | 12 |
| PRAGMA page_size = 8192; VACUUM; | 0 |
| PRAGMA journal_mode=WAL; | 9 |
| PRAGMA cache_size = -20000; | 0 |
| PRAGMA incremental_vacuum; | 0 |
| PRAGMA mmap_size = 33554432; | 0 |
| PRAGMA temp_store = MEMORY; | 0 |
| PRAGMA busy_timeout = 2000 | 0 |

On my web hosting (basic shared hosting also with php 8.3) things look even worse getting 500ms up to multiple seconds:

| Optimization | Time (ms) |
| --- | --- |
| PRAGMA auto_vacuum = incremental; VACUUM; | 418 |
| PRAGMA journal_mode=DELETE; | 574 |
| PRAGMA page_size = 8192; VACUUM; | 0 |
| PRAGMA journal_mode=WAL; | 125 |
| PRAGMA cache_size = -20000; | 0 |
| PRAGMA incremental_vacuum; | 0 |
| PRAGMA mmap_size = 33554432; | 0 |
| PRAGMA temp_store = MEMORY; | 0 |
| PRAGMA busy_timeout = 2000 | 0 |

Note: these measurements arent avarages, just single measurements but I tend to get similar values each time. Also this is not a big database, for testing I just used 5 documents.

Is there any way to disable these optimizations? Are they necessary and/or do they have to run before each search?

Toflar:
Hey, thanks for bringing this to our attention!
First of all, I think in your comparisons, we have to also account for different SQLite versions and different filesystems.
Maybe your shared hosting provider uses an older SQLite version than what you have locally so it's probably pretty normal that a newer version is faster with certain commands because of optimizations that happened inside SQLite?
[...]

PR #181, opened 2025-05-04
Optimize sqlite optimization

daun:
- An attempt at solving #179
- Split optimizations
  - Run startup optimizations on database creation: set page size, enable WAL mode
  - Run connection optimizations on startup: cache, memory, timeout
- Periodically run an incremental vacuum at 2% probability after inserting/updating documents
- Added new tests for the connection settings and the new changes

### DIFF
--- a/src/Internal/Engine.php
+++ b/src/Internal/Engine.php
@@ -206,6 +206,15 @@ class Engine
         return false;
     }
 
+    public function needsVacuum(): bool
+    {
+        if ($this->getIndexInfo()->needsSetup()) {
+            return false;
+        }
+
+        return random_int(0, 100) <= 2;
+    }
+
     public function search(SearchParameters $parameters): SearchResult
     {
         if ($this->getIndexInfo()->needsSetup()) {
--- a/src/Internal/Engine.php
+++ b/src/Internal/Engine.php
@@ -206,15 +206,6 @@ class Engine
         return false;
     }
 
-    public function needsVacuum(): bool
-    {
-        if ($this->getIndexInfo()->needsSetup()) {
-            return false;
-        }
-
-        return random_int(0, 100) <= 2;
-    }
-
     public function search(SearchParameters $parameters): SearchResult
     {
         if ($this->getIndexInfo()->needsSetup()) {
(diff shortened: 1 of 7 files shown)


src/Internal/Engine.php
```diff
@@ -206,6 +206,15 @@ public function needsReindex(): bool
         return false;
     }
 
+    public function needsVacuum(): bool
```
Toflar, line 209: I think we do not need to add this to the `Engine`. The `Indexer` also removes orphans so it's the full responsibility of the `Indexer` to also vacuum. Let's just keep this inside the `Indexer`.

Toflar (commented):
Looks good to me! I just have a nitpick about responsibilities where I think we should move the logic but other than that, I'm happy to merge this!

src/Internal/Engine.php
daun: Sounds good. Will look into it.

daun: Moved the check into the indexer.

daun:
One thing I'm not super happy about is the (non-)testability of the random part. It should work in theory, and I've seen it work with manual testing. Are you okay with leaving it as-is or should we complicate things with some kind of `RandomFactory` we can mock during tests?

daun:
Sure! I've added a `$config->withVacuumProbability(10)` method marked as internal.

Toflar:
So we have a new option now but it's not used in tests? Why adding it then? I'm confused 😄 
Also, I don't think I would add it to the `toArray()` and `fromArray()` if it's purely internal?

daun:
I guess I misunderstood your comment, then. I read it as "no need to complicate the random part for testing, but let's make the number 2 configurable". That was what I attempted here anyway. Let me know if that doesn't track.

Not sure what you mean by "it's not used in tests" — there's a new test that checks the new config methods.

I have a hunch now you did mean to say "let's test the random part properly and add a config to make it easier"?

Toflar:
Yeah I thought you are looking for ways to make the cleanup testable when you commented in https://github.com/loupe-php/loupe/pull/181#issuecomment-2850480998 😊

daun:
Alright, makes more sense now. Will do that.

daun:
Pushed some changes to exclude the probability from the serialized config area, and added tests for 0 and 100 probability.

Toflar (approved):

Toflar:
Great stuff, thanks a lot!

PR #184, opened 2025-05-05
Introducing facets 🥳 

Toflar:
Let's push Loupe to yet another Level. Here we go with support for facets!

Docs still missing but the tests should illustrate. Inspired by Meilisearch again 😎

### DIFF
--- a/src/Internal/Search/Searcher.php
+++ b/src/Internal/Search/Searcher.php
@@ -12,6 +12,7 @@ use Loupe\Loupe\Configuration;
 use Loupe\Loupe\Internal\Engine;
 use Loupe\Loupe\Internal\Filter\Parser;
 use Loupe\Loupe\Internal\Index\IndexInfo;
+use Loupe\Loupe\Internal\LoupeTypes;
 use Loupe\Loupe\Internal\Search\FilterBuilder\FilterBuilder;
 use Loupe\Loupe\Internal\Util;
 use Loupe\Loupe\SearchParameters;
@@ -33,6 +34,8 @@ class Searcher
     public const CTE_TERM_MATCHES_PREFIX = '_cte_term_matches_';
 
     public const DISTANCE_ALIAS = '_distance';
+
+    public const FACET_ALIAS_PREFIX = '_facet_';
 
     public const RELEVANCE_ALIAS = '_relevance';
 
@@ -181,6 +184,7 @@ class Searcher
         $this->selectDocuments();
         $this->searchDocuments($tokens); // First, add the search term CTEs
         $this->filterDocuments($tokens); // Then filter the documents (requires the search term CTEs)
+        $this->addFacets();
         $this->selectTotalHits();
         $this->sortDocuments();
         $this->selectDistance();
@@ -216,7 +220,7 @@ class Searcher
         $totalPages = (int) ceil($totalHits / $this->searchParameters->getHitsPerPage());
         $end = (int) floor(microtime(true) * 1000);
 
-        return new SearchResult(
+        $searchResult = new SearchResult(
             $hits,
             $this->createAnalyzedQuery($tokens),
             $end - $start,
@@ -225,6 +229,8 @@ class Searcher
             $totalPages,
             $totalHits
         );
+
+        return $this->addFacetsToSearchResult($searchResult, $result ?? []);
     }
 
     public function getCTE(string $name): ?Cte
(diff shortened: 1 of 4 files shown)


daun:
🚀

Toflar:
I've added the docs and did a bench. The overhead is minimal in our bench script 😎

Toflar:
Are you sure that's true? Imagine you have moments of time in history and you index the `year` it happened. That would quickly get you a facetDistribution with thousands of individual `years`. I don't think that makes sense?
I think for numeric values, you would want to group them which you should do when indexing, as this is faster anyway. Something like

```php
function groupYearIntoCenturyBlock(int $year): string {
    $start = (int) ($year / 100) * 100;
    $end = $start + 99;
    return "$start-$end";
}

$document['century'] = groupYearIntoCenturyBlock($document['year']);
```

And then you do a facet on the `century` 😇

daun:
Right. Makes sense for large year spans. Assuming a blog archive between 2015 and 2025 where I want to display the years with a count for each — or to stick with the shoe size example between 30 and 50  — how would you handle that? I can imagine storing them as strings in the database, but then I would lose the min/max distribution data.

Toflar:
To be honest, you can achieve that now by just storing them once as a string and once as an integer. KISS 😇

Toflar:
Looks like Meilisearch includes them, though. See https://ecommerce.meilisearch.com for `price` for example. However it seems rounded to integers which makes little sense?

daun:
Their docs have [an example where floats seem to be untouched](https://www.meilisearch.com/docs/learn/filtering_and_sorting/search_with_facet_filters), or stringified:

```json
{
  "facetDistribution": {
    "language": {
      "English": 6,
      "French": 1
    },
    "rating": {
      "2.5": 1
    }
  },
  "facetStats": {
    "rating": {
      "min": 2.5,
      "max": 4.7
    }
  }
}

**Comment by daun** [comment]
Yeah, looks like they round above `1`:

```json
"facetDistribution": {
  "price": {
    "0": 4,
    "0.1": 10,
    "60": 2,
    "65": 1
  }
}
```

Toflar:
Yeah, not very helpful if you ask me. I'd prefer not having this in my result set to improve on performance a little and group the values myself if I need to as illustrated in https://github.com/loupe-php/loupe/pull/184#issuecomment-2854838831.
Thanks for the feedback - Facets, let's go! 🥳

Toflar:
Merged in https://github.com/loupe-php/loupe/commit/9ec8f593fac85a753d631d20adb3cbb6d661a42f, somehow the GitHub UI does not work at the moment.

thePanz:
Nice addition! Great job!

A minor question: from the examples, how do we know in which "stats" or "distribution" a facet will fall into?
Why not having something like:
```
"facets": {
     "age": {
        "type": "stats",
        "values": { ... }
     },
    "department": {
        "type": "distribution",
        "values": { ... } 
    }
}
```

it would allow in the future to add more complex aggregations/facets IMO. WDYT?

notes:
1. I might be biased as a SOLR and ElasticSearch user :sweat_smile: 
2. suggesting this as multi-level aggregations might be a nice feature to add (think of it as a double group-by :-) ). This is supported by Solr/Elasticsearch already, and solves the data fetching for an hypothetical t-shirt shop wanting to display something like  "Give me the facets for the 'size' property, grouped by 'product-id` and for each product-id grouped by 'color' "

Toflar:
Hey Emanuele, long time no hear! 😊 
Loupe is very heavily inspired by Meilisearch, that should answer both your questions 😉 So

> A minor question: from the examples, how do we know in which "stats" or "distribution" a facet will fall into?

String attribute values -> distribution, Numeric attribute values -> stats. Also see the docs.

> it would allow in the future to add more complex aggregations/facets IMO. WDYT?

I don't want to rule out anything but right now, we are capable of doing what Meilisearch is doing and it's probably enough for 99% of all the use cases. If you need more complex aggregations like normal distribution or your multi-level group-by stuff, then Loupe is probably not the right choice 😊

thePanz:
> Hey Emanuele, long time no hear! 

Yeah, life happened in the between :-) 

> String attribute values -> distribution, Numeric attribute values -> stats. Also see the docs.

Yeah, the documentation mentions it, but it would require to check which kind of field was requested as facet.
My fear is that from the available API we can not control which faceting will be applied, and it is decided at runtime.
But I see the aim of the current implementation to keep the complexity low.

> I don't want to rule out anything [..] it's probably enough for 99%

I tend to disagree with the percentage ;-) but I totally see your point of not introducing a feature which would require implementation+maintenance for a minor use-case.

Thanks again for the great tool! :+1:

PR #187, opened 2025-05-08
Remove ellipses from bullet points in README.md

nz-chris:
In my opinion, scanning the list is easier and a more pleasant experience without them!


Toflar:
Thanks!

Issue #13, opened 2023-08-04
Support for offset and limit


alexander-schranz:
Currently it is build for only paginated views. But it would be great if also offset and limits would work. Some usecase for this e.g.:

Display first the Top 10 matches but have a button showAll which will then `offset: 10` without any `limit`.

My suggestion is to use in `SearchParameters` `limit` and `offset` but have 2 methods e.g.:

 - `withPagination(int $page, int $hitsPerPage);`
 - `withLimitOffset(?int $limit, int $offset);`

Toflar:
I have yet to see a use case for `offset` and `limit`. To me, `page` and `hitsPerPage` is so much more intuitive.

I also don't think that "without any limit" exists in reality. There is always a limit. Either you define a limit by configuring it or - if we would allow to configure it to `null` or `PHP_INT_MAX` -  Loupe crashes because there's too much data to fetch. This is also a limit. And crashing is probably never what you want 😄 
That's why Loupe does not allow a value higher than `1000` at the moment.

I think it's better to force users to think about a limit rather than building an application that would work without a limit. Again, no application can work without a limit, it's a logical problem, not a Loupe configuration problem.

By the way: MeiliSearch also has that limit https://www.meilisearch.com/docs/reference/api/settings#body-5

alexander-schranz:
I understand the point that there is always a limit and not a null or PHP_INT_MAX used, for my case the issue is more about support for `offset`.

The offset limit has other advantages you only can first show `2` elements and then on a more load button load the next 10 elements (offset: 2, limit: 10) this is with a normal pagination not possible.

Toflar:
Yeah I don't want to eliminate the possiblity of having `offset` and `limit` making it into Loupe. I just don't need it at the moment and anybody is free to work on that. I think we have to pay extra attention to what happens if it's used in combination with `hitsPerPage` and `page`. I know MeiliSearch also allows both so one just has to take precedence or we have to throw an exception. Not sure yet :)

PR #188, opened 2025-05-09
Added support for offset and limit

Toflar:
Fixes https://github.com/loupe-php/loupe/issues/13.

### DIFF
--- a/src/Internal/Search/Searcher.php
+++ b/src/Internal/Search/Searcher.php
@@ -217,15 +217,17 @@ class Searcher
         }
 
         $totalHits = $result['totalHits'] ?? 0;
-        $totalPages = (int) ceil($totalHits / $this->searchParameters->getHitsPerPage());
+        $hitsPerPage = $this->queryBuilder->getMaxResults() ?? 0;
+        $totalPages = $hitsPerPage === 0 ? 0 : ((int) ceil($totalHits / $hitsPerPage));
+        $currentPage = $hitsPerPage === 0 ? 0 : ((int) floor($this->queryBuilder->getFirstResult() / $hitsPerPage) + 1);
         $end = (int) floor(microtime(true) * 1000);
 
         $searchResult = new SearchResult(
             $hits,
             $this->createAnalyzedQuery($tokens),
             $end - $start,
-            $this->searchParameters->getHitsPerPage(),
-            $this->searchParameters->getPage(),
+            $hitsPerPage,
+            $currentPage,
             $totalPages,
             $totalHits
         );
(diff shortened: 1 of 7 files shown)


Toflar:
@alexander-schranz would you mind checking if this works correctly and you can get rid of all the special handling for Loupe in SEAL?

src/Internal/Search/Searcher.php
```diff
@@ -967,12 +969,20 @@ private function formatHit(array &$hit, TokenCollection $queryTerms): void
...
-        $maxPageOffset = $maxTotalHits - $hitsPerPage;
 
-        $this->queryBuilder->setFirstResult(min($pageOffset, $maxPageOffset));
-        $this->queryBuilder->setMaxResults($hitsPerPage);
+        $offset = $this->searchParameters->getOffset();
+        $limit = $this->searchParameters->getLimit();
+
+        if ($this->searchParameters->getHitsPerPage() !== null || $this->searchParameters->getPage() !== null) {
+            $limit = $this->searchParameters->getHitsPerPage() ?? SearchParameters::MAX_LIMIT;
+            $offset = (($this->searchParameters->getPage() ?? 1) - 1) * $limit;
```
alexander-schranz, line 978: I accidently did:

```php
$searchParameters = $searchParameters->withHitsPerPage($limit);
// some more code
$searchParameters = $searchParameters->withLimit($limit);
$searchParameters = $searchParameters->withOffset($offset);
```

but the offset is because of `withHitsPerPage`, never keep in mind.

Maybe on `withLimit` or `withOffset` set hits per page and page to null.

Maybe we could also remove withPage, withHitsPerPage, withLimit and withLimit and just have: `withPagination($hitsPerPage, $page)`, `withLimitAndOffset($limit, $offset)`.

and withPagination also just set limit and offset values.

alexander-schranz (approved):
SEAL tests work as expected but requires following changes to support the new feature:

```diff
diff --git a/packages/seal-loupe-adapter/src/LoupeSearcher.php b/packages/seal-loupe-adapter/src/LoupeSearcher.php
index 72301b3..64d3064 100644
--- a/packages/seal-loupe-adapter/src/LoupeSearcher.php
+++ b/packages/seal-loupe-adapter/src/LoupeSearcher.php
@@ -80,7 +80,11 @@ final class LoupeSearcher implements SearcherInterface
         }

         if ($search->limit) {
-            $searchParameters = $searchParameters->withHitsPerPage($search->limit);
+            $searchParameters = $searchParameters->withLimit($search->limit);
+        }
+
+        if ($search->offset) {
+            $searchParameters = $searchParameters->withOffset($search->offset);
         }

         if ([] !== $search->highlightFields) {
@@ -91,12 +95,6 @@ final class LoupeSearcher implements SearcherInterface
             );
         }

-        if ($search->offset && $search->limit && 0 === ($search->offset % $search->limit)) {
-            $searchParameters = $searchParameters->withPage((int) (($search->offset / $search->limit) + 1));
-        } elseif (null !== $search->limit && 0 !== $search->offset) {
-            throw new \RuntimeException('None paginated limit and offset not supported. See https://github.com/loupe-php/loupe/issues/13');
-        }
-
         $sorts = [];
         foreach ($search->sortBys as $field => $direction) {
             $sorts[] = $this->loupeHelper->formatField($field) . ':' . $direction;
```

For adding then support beside the new tag of loupe also new requirement loupe/matcher is required to be released.

src/Internal/Search/Searcher.php
Toflar: This is sort of the expected behavior. We are currently very close to the Meilisearch API. And there you can do exactly the same. If you pass

```json
{
    "hitsPerPage": 15,
    "offset": 10,
    "limit": 20
}
```

according to the docs, `offset` and `limit` are just going to be ignored. And because we're close to MS, we should also leave "offset" and "limit" be two separate methods.

Toflar:
Neat, thanks for checking! I'm merging this then 😊 I will likely release a new version of Loupe this month.

Issue #174, opened 2025-04-22
Issue with maxTotalHits


Toflar:
I have a problem now because I used Loupe to iterate over sets of 5'000 objects and more but due to https://github.com/loupe-php/loupe/pull/172 I can't do this anymore now as the `totalHits` will now always be `1000` and thus the pagination thinks it's done :/

How would you now loop over all the objects in the index @daun ? Any ideas?

daun:
That was the idea — max total hits applies regardless of pagination. But it should be configurable. Does the below work?

```php
$config = \Loupe\Loupe\Configuration::create()->withMaxTotalHits(5000);
```

Toflar:
Yeah sure, this would work but I'm not sure this is what I want to do. Afaik scraping the entire index is the problem for Meilisearch which is designed to expose the API publicly. This does never happen for Loupe anyway, it always has to go via your own code if you want to expose it to the public.
But yeah, kinda wondering what to do now. I want to loop over all documents in the index - I need that feature...

Toflar:
Exactly. In my case I'm generating some sitemap data for all the documents in the index using a cronjob. So for now I iterated increasing the page parameter on every request. The `Indexer` should remain internal - I guess it's not wrong to have that on `Loupe` itself, just wondering how this should be named. How do other search engines deal with this anyway? 🤣

daun:
Maybe you're right and the `maxTotalHits` applying as a hard limit doesn't quite make sense? I liked the parity with Meilisearch, but we could also switch to a custom param that only applies as a limit of documents per term, as with your const MAX_TERM_HITS or similar. In that case, a good old search for `*` would work as before.

Toflar:
It does make sense imho to limit the term hits because that could cause Loupe to break. But browsing is not something that is of any issue to us, imho. You can limit the results per page using `withHitsPerPage()` and this cannot exceed `1000` either anyway. That's also enforced on `SearchParameters` atm.

Toflar:
I'm not against `browseObjects()` but why taking someone the opportunity to loop over all the documents and e.g. apply a sort order 🤷‍♂

daun:
I think re-using `hitsPerPage` for the per-term document limit could be problematic. When using a page size of, say, 100, this would mean that most relevance scores are basically useless as we only ever consider the first 100 documents per term. So I guess we'd need a separate parameter here.

Also, a `browseObjects` API could of course have a sort param as well.

So I see two options here:

[...]

PR #191, opened 2025-05-12
Added browse API

Toflar:
Should fix https://github.com/loupe-php/loupe/issues/174.

Please review @daun 😊


daun:
Looks good! I like the abstraction :)

PR #205, opened 2025-06-20
Re-use existing match information

Toflar:
Just a draft, let's see what @daun thinks 😊 

Requires https://github.com/loupe-php/matcher/pull/9.

### DIFF
--- a/src/Internal/Search/Searcher.php
+++ b/src/Internal/Search/Searcher.php
@@ -988,9 +988,16 @@ class Searcher
             return new FormatterResult($value, new TokenCollection());
         }
 
-        // TODO: pass on matchesPositionInfo somehow to the formatter so that it can re-use the positional info and does
-        // not tokenize the $value again
-        return $this->engine->getFormatter()->format($value, $queryTerms, $attributeOptions);
+        $matches = new TokenCollection();
+
+        // TODO: must use stop words and include them here
+        foreach ($this->engine->getTokenizer()->tokenize($value)->all() as $i => $token) {
+            if (\in_array($i + 1, $matchPositionInfo[$attribute], true)) {
+                $matches->add($token);
+            }
+        }
+
+        return $this->engine->getFormatter()->format($value, $queryTerms, $attributeOptions, $matches);
     }
 
     /**


daun:
Nice :) Out of curiosity — how does this speed things up technically? By only passing in the tokens that we already know have matched? Looks like it's still tokenizing things during formatting, and just filtering them down?

Toflar:
Tokenizing is not the problem. Matching is 😊 Imagine you have a huge text and you now want to highlight `foo` in that text. Currently, we tokenize this huge text and then foreach over every single token to check if this matches `foo`. Although we already know where in this huge text `foo` matches.

daun:
Right! Yeah, that sounds like huge savings in processing time.

daun:
@toflar From a quick look, I think the tests are failing because the match positions aren't compiled correctly for array values. When logging inside `formatHit`, I'm getting this for the array values:

```php
["_match_position_info_0" => "genres:2,genres:2"]
```

Which is for this result:

```php
[
  "id" => 12,
  "title" => "Finding Nemo",
  "overview" => "Nemo, an adventurous young clownfish, is unexpectedly taken from his Great Barrier Reef home to a dentist's office aquarium. It's up to his worrisome father Marlin and a friendly but forgetful fish Dory to bring Nemo home -- meeting vegetarian sharks, surfer dude turtles, hypnotic jellyfish, hungry seagulls, and more along the way."
  "genres" => [
    "Animation",
    "Family"
  ]
]
```

Not sure what `genres:2` means here, and how it relates to the matches. It's either that the index of the position is wrong, or that it's missing a nested array key. Or both?

daun:
Could it be that we don't have the required information from the database? Meaning the array index and match position of each individual match? That would be necessary for formatting.

daun:
Took a bit of a dive. The `LoupeTypes::convertToString` method will turn the array of strings from the example into something like this: `0.Animation.1.Family`, which would yield the expected match position `2` for "Animation". I don't think there's a good way of decoding that part and getting exact match positions for highlighting multidimensional arrays. At least not without a major rewrite of how arrays are handled.

Mmh. Maybe for now just highlight arrays like before, by tokenizing and matching on the fly? We would get the speed gains for top-level string fields at least, and figure out arrays later.

Toflar:
Honestly, I never really understood highlighting in arrays. Personally, I have no project where I need to highlight anything other than just a string (no matter the search engine used).

> Mmh. Maybe for now just highlight arrays like before, by tokenizing and matching on the fly? We would get the speed gains for top-level string fields at least, and figure out arrays later.

I agree. We might add an info in the `Performance` section or so. Any hint/PR on what needs to be adjusted then?

daun:
We'd just need to adjust the formatting part to not pass in the matches when formatting array values. I've created a PR with the missing pieces for tests to pass: #214

PR #211, opened 2025-07-03
Always sort null and empty values last

Toflar:
I think we're doing this wrong currently because when I sort by e.g. `title` alphabetically, then all the `null` and `''` (empty string) values come first.
I think this never makes sense from a search engine's perspective. They should always come last, no matter if I order `asc` or `desc`. If I don't want to have them in the result set at all, then I can `filter` them out but when I just search for `foobar` and want to order them alphabetically then those without a `title` should never appear first no matter if descending or ascending.

Or an other example for a `rating` property: If I order results descending, then I want the highest rated documents and never null values first. And if I order ascending, then I want the lowest rated documents, never null values first.
Again, if they should not be part of the result set at all then I filter them out but if I do not want to filter but just sort, then I never care about those without a rating at all. 

Wdyt @daun @alexander-schranz?

### DIFF
--- a/src/Internal/Search/Sorting/AbstractSorter.php
+++ b/src/Internal/Search/Sorting/AbstractSorter.php
@@ -56,14 +56,15 @@ abstract class AbstractSorter
         $alias = $cteName . '.sort_order';
 
         // Because of how Loupe works (SQLite's loosely typed system) we need to always ensure that null and empty values
-        // are ordered (in the opposite way) first, so that they then match with the regular sorting
+        // are ordered ascending first.
+        // Null and empty values should always come last for Loupe (or generally speaking for any search engine probably).
         $searcher->getQueryBuilder()->addOrderBy(
             Operator::Equals->buildSql(
                 $engine->getConnection(),
                 $alias,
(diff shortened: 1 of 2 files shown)


daun:
The reasoning sounds good!

From what I remember from Postgres, nulls are larger than non-nulls, so in asc sort they come last and in desc sort they come first. But the sort value of null can be adjusted in the query itself.

I like putting them always last, especially in a search engine context.

daun:
In that case putting empty values last sounds great!

Toflar:
I think this is fine, so I will merge this.
For reference:

- In Elasticsearch you can configure the behavior (because in ES you can configure everything): https://www.elastic.co/docs/reference/elasticsearch/rest-apis/sort-search-results#_missing_values
- In MeiliSearch, missing values are placed last, so just like we will do now (https://specs.meilisearch.dev/specifications/text/0055-sort.html#_1-functional-specification)

alexander-schranz:
@Toflar makes totally sense from my point of view. We maybe should in SEAL change where possible (not all are at first check e.g. Redis) the behaviour also to missing last.

PR #212, opened 2025-07-09
Ignore invalid attributes during setup

Toflar:


Toflar:
Stupid me, this was not the right fix. https://github.com/loupe-php/loupe/pull/213 is.

PR #222, opened 2025-07-25
Increase the pack size for the stored state set

Toflar:
Depending on how you configure the typo tolerance alphabet size and index length, you can easily end up having states that exceed the 32bit maximum of 2^32 - 1. So let's switch to 64 bits here.

Wdyt @ausi?

### DIFF
--- a/src/Internal/Engine.php
+++ b/src/Internal/Engine.php
@@ -33,7 +33,7 @@ use Toflar\StateSetIndex\StateSetIndex;
 
 class Engine
 {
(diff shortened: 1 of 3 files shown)


Toflar:
Huh, how does that make the tests fail now - were they wrong before?

ausi:
We could calculate the maximum possible state based on alphabet size and index length and select the appropriate integer size based on that. But depending on the size of this file in real world use-cases this might be an unnecessary premature optimization.

Toflar:
We're talking about like 420kb or so when using the 32k movies database.

PR #227, opened 2025-08-05
Fix and validate document schema before indexing

Toflar:
At the moment, we're validating the schema of every document while indexing. This can cause document schema updates within the transaction which will in turn then lock the database. So you cannot search while indexing is taking place (only if there are schema changes).

This PR changes this so that we're validating and fixing schema stuff first and only then we'll start indexing.

/cc @daun

### DIFF
--- a/src/Internal/Index/Indexer.php
+++ b/src/Internal/Index/Indexer.php
@@ -81,7 +81,7 @@ class Indexer
 
         try {
             $this->engine->getConnection()
-                ->transactional(function () use ($indexInfo, $documents, &$successfulCount, &$documentExceptions) {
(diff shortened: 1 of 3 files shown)


daun:
Looks good! Am I correct in assuming that chunking document imports has always somewhat mitigated this by splitting documents across multiple transactions? E.g. importing 1000 documents as 10 batches of 100 documents in a row. That would explain why I never ran into this on production systems.

Toflar:
It does mitigate the problem, yes. But mostly, if you make sure the schema is identical from the very beginning (so all your documents have the exact same structure), you won't have the problem. It only happens if you have in-between schema changes (which is bad practice anyway but still 😊 ).

Issue #232, opened 2025-10-05
Exception when trying to use pdo_sqlite


manuxi:
Tried to use loupe using pdo_sqlite only throws the followinge exception:
`{
    "code": 0,
    "message": "Attempted to load class \"SQLite3\" from the global namespace.\nDid you forget a \"use\" statement?",
    "errors": [
[...]

PR #233, opened 2025-10-27
Run tests without sqlite3 and pdo_sqlite extensions

Toflar:
Fixes https://github.com/loupe-php/loupe/issues/232.

Run the CI without `sqlite3` and `pdo_sqlite` in the matrix so that we can cover this case as well.

### DIFF
--- a/src/LoupeFactory.php
+++ b/src/LoupeFactory.php
@@ -84,8 +84,8 @@ final class LoupeFactory implements LoupeFactoryInterface
 
         $sqliteVersion = match (true) {
             \is_callable([$connection, 'getServerVersion']) => $connection->getServerVersion(), // @phpstan-ignore function.alreadyNarrowedType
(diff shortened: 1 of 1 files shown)


src/LoupeFactory.php
alexander-schranz: @Toflar looks for me now that this exception is not longer thrown? Think moving it into the try would output the `InvalidConfigurationException` correctly?

```diff
       } catch (\Throwable) {
-           if (!class_exists(\PDO::class)) {
-               throw new \RuntimeException('pdo_sqlite not installed.');
-           }

            try {
+               if (!class_exists(\PDO::class)) {
+                   throw new \RuntimeException('pdo_sqlite not installed.');
+               }

                $connection = DriverManager::getConnection(
                    $dsnParser->parse('pdo-sqlite://' . $dsnPart),
                    $this->getDbalConfiguration($configuration)
                );
            } catch (\Throwable) {
                // Noop
            }
        }
```

Or did I miss something?

alexander-schranz: or #235 😄

PR #238, opened 2025-11-06
Functional tests to illustrate concurrency issues

Toflar:
The current version 0.12 mainly suffers from two key issues that I would like to improve in v0.13. In my opinion, these are also the things that still keep me from calling this a "stable software" and thus releasing version 1.0.

### 1st key issue: Concurrency handling

SQLite is great. Its flexibility and simplicity enabled a project like Loupe in the first place. However, there is one major issue: There cannot be multiple, concurrent writers. There can only ever be one process writing to an SQLite database. While in normal environments this is not an issue, it is for Loupe. Normally, writing to the database happens quickly in very short transactions. So if your writes take milliseconds like they normally do, you can have hundreds of concurrent writing processes. However, indexing thousands of documents can take pretty long. That means that one process can block another one or in worst case, cause the other one to abort due to `database locked` if it has to wait longer than `busy_timeout` which is currently set to 5 seconds. That's also why we currently index one document after the other in order to hopefully stay within the 5 seconds. This, however, is not a long term solution. Bigger documents (= taking longer to index) combined with a growing number of concurrent processes can still cause Loupe to fail.

### 2nd key issue: too much file I/O

Having this current workaround of one document per transaction **drastically** increases file I/O. 10,000 inserts = ~10,000 journal writes + ~10,000 fsyncs (or at least a large number of syncs depending on mode). Also, currently we do loads of UPSERT queries. So every single term or attribute value is UPSERTed in its own query. We do not batch insert data which also has a huge negative impact on the index speed.

### What does this PR do?

This PR is the first step towards v0.13. It provides the functional tests illustrating all the current issues (and also potential new issues) so we can do test-driven software development to fix all the failing tests. It includes 3 main tests:

1. Ensure correct order:
When `worker-1` is started before `worker-2`, it must be properly queued. If both want to write document ID `the-one-in-question`, `worker-2` **must** win over `worker-1` because it was triggered after. Currently if `worker-1` takes 1 minute to index, a concurrently running `worker-2` could index `the-one-in-question` before `worker-1` makes it to that document. This will cause `worker-1` to write **after** and thus win over `worker-2` which is wrong. 
2. Tolerate failing processes:
We will have to implement some queuing logic in some way. I plan to implement some "ticket" architecture. Meaning that every modifying process retrieves a ticket first, ensuring its position in the queue. It can then process whatever it wants once it is its turn. Once the process is done, it would update a second database (cannot use `loupe.db` because this one is potentially locked) to inform other processes that it's done and the next ticket can do whatever it wants to do. That should fix 1). However, if a process is killed for some reason, this would cause its ticket to be never acknowledged, hence causing all other processes to be blocked forever. We have to make this fail-proof which is covered in those functional tests as well now.
3. Tolerate bigger documents
All of this should allow not running into errors when documents take longer than 5 seconds to index. This currently fails which this PR also covers.

A nice side effect of those tests is that we can also compare the **indexing speed** between the current implementation and whatever solution I come up with.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,17 +33,19 @@
         "loupe/matcher": "^0.2"
     },
     "require-dev": {
-        "symfony/var-dumper": "^7.0",
-        "symfony/filesystem": "^7.0",
-        "symplify/easy-coding-standard": "^12.6",
-        "symfony/finder": "^7.0",
+        "phpstan/phpstan": "^2.0",
         "phpunit/phpunit": "^10.5",
-        "phpstan/phpstan": "^2.0"
+        "symfony/filesystem": "^7.0",
+        "symfony/finder": "^7.0",
+        "symfony/process": "^7.0",
+        "symfony/var-dumper": "^7.0",
+        "symplify/easy-coding-standard": "^12.6"
     },
     "scripts": {
         "tests": "@php vendor/bin/phpunit",
         "unit-tests": "@php vendor/bin/phpunit --testsuite=unit",
         "functional-tests": "@php vendor/bin/phpunit --testsuite=functional",
+        "slow-tests": "@php vendor/bin/phpunit --testsuite=slow",
         "cs-fixer": "@php vendor/bin/ecs check --fix",
         "ci-cs-fixer": "@php vendor/bin/ecs check",
         "phpstan": "@php vendor/bin/phpstan analyse"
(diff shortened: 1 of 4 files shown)


Toflar:
Merging this into `develop` now because fixing this is what v0.13 will be all about 😊

PR #239, opened 2025-11-07
Split preparing and writing and introduce a ticket handler to fix concurrency issues

Toflar:
I have been working on this PR all week and it should address the concurrency issues outlined and described in https://github.com/loupe-php/loupe/pull/238.

Notable changes vs the current 0.12:

* Split up preparing the documents (tokenization, stemming, language detection, etc.) and actually writing. This enables better performance in concurrent environments because if one process has to wait for another to complete, it can still already prepare all the documents.
* Introduced a `TicketHandler`. Every writing operation (mainly adding and deleting documents) now first claims a ticket (separate database) in order to ensure its position in the queue. Then it prepares the documents and once its ready and its spot is available, it will index.
* Because we now are guaranteed to be the only writer, we can wrap it all into one exclusive transaction again, meaning we can insert e.g. 1000 documents within one transaction instead of 1000.
* Dropped `IndexResult`, it's all regular exception handling like for all the other methods too.

This now fixes all failing concurrency tests in the `develop` branch and it also already brings some nice improvements when running `bin/bench/index.php` with the 30k movies:

| Branch | Index results                        |
|--------|--------------------------------------|
| main   | Indexed in 237.90 s using 98.50 MiB  |
| PR     | Indexed in 124.52 s using 246.50 MiB |

So indexing time is **cut by half**. The memory usage is way higher because we prepare all documents before writing them now. So they are all kept in memory of course. However, this can be influenced by the users themselves. Instead of calling `addDocuments()` with 10000 documents, they can split them into chunks of 1000 for example.

Tracking file I/O is harder as I'm not really super familiar with it but I just used `fs_usage` on MacOS and tried filtering everything that writes to either `loupe.db` or `tickets.db` (new in this PR) like so:

```
sudo fs_usage -f filesys -w | grep -E 'loupe\.db(-wal)?|tickets\.db(-wal)?' | grep -E 'Wr(Data|Meta)|p?write|fsync' > file_access.log
```

I started this command and then ran `php bin/bench/index.php` on the 30k movies. After that, I counted the entries in the `file_access.log` file like so: `cat file_access.log | wc -l`.
Again, not sure how accurate but I think it's still pretty telling 🤯 

| Branch | File I/O log lines                        |
|--------|--------------------------------------|
| main   | 3 934 607  |
| PR     | 51 421 |

This PR now also opens the possiblity for yet another optimization to follow: bulk inserts! (which should also affect memory usage positively) 🥳

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Note that some implementation details (e.g. libraries used) referenced in this b
 Performance depends on many factors but here are some ballpark numbers based on indexing the 
 [~32k movies fixture][MeiliSearch_Movies] provided by MeiliSearch.
 
-* **Indexing** will take around **230 seconds** (~140 documents per second)
+* **Indexing** will take around **130 seconds** (~250 documents per second)
 * **Querying** for `Amakin Dkywalker` with typo tolerance and relevance ranking takes about **100 ms**
 
 Note that anything above 50k documents is probably not a use case for Loupe. You can run your own benchmarks 


Toflar:
> The memory tradeoff is great as batching can always happen on the consumer side to optimize.

Regarding the memory, most of it is related to prepared statements so I'm fairly sure these numbers will change/can be optimized in follow-up PRs.

> One question I can't seem to find the answer to myself scanning the changes: when an exception occurs during indexing, does this now abort the whole batch? So if, say, the 3rd of 100 documents has an error, do the other 97 get indexed anyway?

Correct. However, as stated in the initial PR description, I want to give bulk inserts a shot. So we'll have one query for many documents which should - hopefully - have another positive effect on performance and file I/O.
And to be honest, I'm not sure if we can then handle this case properly as we'd have one query for e.g. 100 documents. It might be doable with some `ON CONFLICT() IGNORE` stuff but I'd say, let's see after I've played around with bulk inserts. If you then want to give it a shot, I wouldn't be opposed of course 😉 

Merging this then 😊

PR #244, opened 2025-11-14
Skip identical documents

Toflar:
After a very long time of debugging and learning about SQLite behavior with different drivers in PHP, I've come to the conclusion that I would like to drop the support for `sqlite3` and henceforward require `pdo_sqlite`.

Basically, there are quite a few quirks in `sqlite3` in PHP. See https://github.com/php/php-src/pull/5204#issuecomment-810095573. There was no way for me to get this working with it, the information coming from the database is swallowed by it :/

So compared to the current `v0.12` this would make the current state of `develop` change two platform requirements:

* `sqlite3` is not supported anymore, must have `pdo_sqlite`
* minimum required sqlite version increased from `3.16.0` to `3.35.0`

However, I'm pretty confident that this is a good decision because the performance improvements are just massive. And this PR provides even more.

Run `php bin/bench/index.php --no-delete` 2 times. So that the second run would only update all the documents. Previously, both runs would take about `32 seconds` (yes, we've gotten even faster in the meantime...).

Now the second run will take just `5 seconds` because there's literally nothing to do 😊 

Wdyt @daun @alexander-schranz? Are the changed requirements okay? I really don't feel like investing even more hours and days just to provide some polyfill for older systems. Not opposed if anyone wants to do it but I don't really see it happening for me atm...


alexander-schranz:
I think its even better this way as we can require pdo sqlite extension now in the `composer.json` which should work then great with [PIE](https://github.com/php/pie).

PR #267, opened 2026-01-07
Support compound word splitting

Toflar:
I guess this would mark another milestone in the history of Loupe. For now especially for German and English contents 😎 

I currently do not have a lot of German documents at hand but maybe I can get some performance stats from @alexander-schranz or @daun here? 😉


daun:
I'll try this on a production database with lots of German content this week or beginning of next week and report back about performance. Amazing :)

Toflar:
And of course this is only that expensive because there are almost never matches when applying English to a German dictionary. So it's almost always `log2(140k) ≈ 17 comparisons` to find the match. This will of course perform much better if there are earlier matches.

daun:
Just tried this on a production database (~ 15K content entries). The indexing works fine and is reasonably fast.

However I can't get to the actual searching because it's running out of memory in `nitotm/efficient-language-detector/
resources/ngrams/small.php`. Has something changed about how that library is used during search?

Toflar:
> So here are some numbers. Not too much difference. None, actually. Will try another dataset later.

2 things that make me scratch my head:

1. How many terms in general are we talking? I mean, does decomposition even work for you? (aka is the language detection working and thus decomposition happens) - because those stats sound a bit too good to be true as they are basically identical which I can only explain by the fact that in your case maybe decompositon doesn't happen at all or only for a very little amount of terms?
2. An increase in memory is expected due to the dictionary being loaded in to memory. However, for German this should only be about 15 MB. Is it because Dutch is also detected and so both dictionaries are loaded? 🤔 And also, how do you get to 240 MB (before) in the first place? 😄

Toflar:
Found some language detection issue which caused `nl` to be loaded in my case even though Dutch was nowhere near the correct language 😉 See: https://github.com/loupe-php/loupe/releases/tag/0.13.7
Hence, rebased this PR again. But shouldn't affect your tests (other than maybe not loading a dictionary that's not actually needed).

daun:
1. Yes, composition is working well. Tested it and it finds words now it didn't find before ("Webstuhl" → Stuhl, "Filmvorführung" → Vorführung), so it definitely works.
2. Should have said that as well: I'm doing the tests inside a CMS environment because that's how I can access the data. I don't have a raw JSON dump. Most of that memory is Laravel & the CMS bootstrapping, so the absolute numbers don't matter much.

~~One suspicion I have is that the language detector marks most documents as English since there are so many loanwords (Performance, Event, etc) that just trigger false positives even though the document is in German. Is there any way of debugging the detected language?~~ Oh, just saw the latest patch release which should sort this out :)

Toflar:
With the latest 2 byte prefix bucket search this is down from `43.36%` to `32%`.

Toflar:
I noticed that we'll also need that for English as English also has a lot of compound words (e.g. `toothbrush`) 😄 I ran the benchmark on the `movies.json` with our 32k movies and saw an increase of like 30% which I think is totally fine.
Improving quality means improving the dictionary which will also improve the performance (less noise -> less splits -> less work). So I guess the discussion should be continued in the respective PR: https://github.com/loupe-php/matcher/pull/14

daun:
I think the failing tests in this one should resolve themselves once we merge `main` into `develop` run the tests here again.

daun:
🎉

daun:
@Toflar This one got auto-closed when the `develop` branch was deleted. Should I rebase onto `main`?

Toflar:
My goodness - had to change the state and target branch of this PR via GitHub API 🤦

daun:
Hehe. We could have just opened a new one, but nice to keep the pr comments around.

@Toflar  Can you manually retrigger the tests? I've updated the matcher branch with the latest changes from its main branch, so re-running should theoretically fix most test errors. There's one new failing test locally which i can't make sense of, but it's a start :)

daun:
Fascinating. So I found the issue why that single test fails.

The new umlaut exact-match boost from #288 does not work on this branch because we now use a German normalizer when decompounding – but only for the document, not for the query.

Documents are tokenized with the German locale (LanguageDetector::detectForDocument does not require reliablity), but queries are tokenized without a locale (LanguageDetector::detectForString requires reliability). That asymmetry causes the document's `müller` to become `mueller` while the query's `müller` becomes `muller`. Document and query produce different terms from the same input, and the `wasFolded` exact-match logic can't apply.

How to reproduce:

```
vendor/bin/phpunit --filter testRelevanceAndRankingScoreForNormalizedSpelling
```

Why does main work? The current latest matcher (0.2.4) has no locale-specific normalizer, so `müller` becomes `muller` regardless of locale. Document and query produce the same term.

I figure we have two options. Both don't sound great or reliable. Can you think of anything else?

1. Make query tokenization use the same "best language" fallback as document tokenization. That could be rather unreliable and lead to surprises when no good language guess can be made.
2. Drop the umlaut folding from `GermanNormalizer` (keep ASCII folding as the only path) so docs and queries can't diverge regardless of detected locale. That would break proper decomposition I reckon.
3. ???

Toflar:
Yeah that's what happens with language detection. Sometimes it can reliably detect the language and sometimes it can't. Looks like `thomas müller` is not German enough for the ngram based detection system. For us it's probably as German as it can get 😉 Anyway, I don't think this is an issue in the logic because I think it's correct that we try to normalize as close to the detected language as we can. 
Fixed tests by just forcing `de` in those examples.

daun:
Nice, that sounds very pragmatic.

Toflar:
@daun guess this is ready then. The changes in the tests are because of v0.3 of `loupe/matcher`, right? Should probably outsource that in a separate PR because not related to the rest?

daun:
@Toflar Looks like it. If you see a clean way of splitting things, that sounds like a good plan. I tried and failed recently. Life shouldn't be hard so I think it's fine if they don't split perfectly but we merge them around the same time anyway :)

Toflar:
See https://github.com/loupe-php/loupe/pull/320 - I actually found a better solution for that `thomas müller` problem than hardcoding the language to German. I think it is just wrong to assume that the detection has to be `reliable` for a search query because search queries very often consist of just one term so most of the time, either the number of ngrams or the reliability are just too little/too low - meaning that it would almost always split according to `null` (aka no language specific rules). However, I think it's always better to split according to the best match even if not reliable - only for the query case 😊

PR #272, opened 2026-02-18
Fix unconditional DISTINCT in filter/sort-only queries

### LINKED ISSUE #271: Avoid unnecessary DISTINCT in filter/sort-only queries
While investigating slower-than-expected query performance on a dashboard powered by Loupe, I noticed that `Searcher::filterDocuments()` applies `->distinct()` unconditionally to the matches query builder — even in cases where there is no search query string (i.e., filter-only or sort-only operations).

https://github.com/loupe-php/loupe/blob/eaec295de865f29cb8c8e86111f6faaaefb7d0d7/src/Internal/Search/Searcher.php#L1016

This appears to have a measurable performance impact depending on query type

**Observations**

| Scenario | With `distinct()` | Without `distinct()` |
|---|---|---|
| Filter/sort only (no query string) | Slower | **Faster** |
| With query string | **Faster** | Slower |

**Query Plan Analysis**
Using `EXPLAIN QUERY PLAN` shows that `DISTINCT` forces SQLite to materialize the CTE and build additional indexes for filter/sort-only queries:

```bash
# Query plan with $qbMatches->distinct() (filter/sort only)
CO-ROUTINE (subquery-5)
  MATERIALIZE _cte_matches
    SCAN d USING COVERING INDEX IDX_A2B07288A8EBE516
  MATERIALIZE order_release_date
    SCAN _cte_matches
    SEARCH d USING INTEGER PRIMARY KEY (rowid=?)
    USE TEMP B-TREE FOR GROUP BY
  SCAN order_release_date
  BLOOM FILTER ON _cte_matches (document_id=?)
  SEARCH _cte_matches USING AUTOMATIC COVERING INDEX (document_id=?)
  SEARCH d USING INTEGER PRIMARY KEY (rowid=?)
SCAN (subquery-5)
USE TEMP B-TREE FOR ORDER BY
```

```bash
# Query plan example for a filter/sort query without $qbMatches->distinct()
CO-ROUTINE (subquery-5)
  CO-ROUTINE order_release_date
    SCAN d
    SEARCH d USING INTEGER PRIMARY KEY (rowid=?)
  SCAN order_release_date
  SEARCH d USING INTEGER PRIMARY KEY (rowid=?)
  SEARCH d USING INTEGER PRIMARY KEY (rowid=?)
SCAN (subquery-5)
USE TEMP B-TREE FOR ORDER BY
```

I can't say with 100% certainty that CTE materialization is the sole cause of the slowdown, but the benchmarks below suggest it could be a significant factor. I might be off track, but the overhead from `$qbMatches->distinct()` when there's no search term — and thus no risk of duplicate matches (?) — seems potentially unnecessary.

**Benchmarks (SQLite 3.51.0 Apple M1)**

```bash
# Sort only with DISTINCT
for i in {1..5}; do php bin/bench/sort.php; done

Searched in 69.43 ms using 6.00 MiB
Searched in 64.78 ms using 6.00 MiB
Searched in 68.12 ms using 6.00 MiB
Searched in 66.97 ms using 6.00 MiB
Searched in 65.58 ms using 6.00 MiB

# Sort only without DISTINCT
for i in {1..5}; do php bin/bench/sort.php; done

Searched in 37.91 ms using 6.00 MiB
Searched in 38.36 ms using 6.00 MiB
Searched in 37.50 ms using 6.00 MiB
Searched in 39.88 ms using 6.00 MiB
Searched in 37.91 ms using 6.00 MiB

# Search query without DISTINCT
for i in {1..5}; do php bin/bench/search.php 'star wrs'; done

Searched in 223.07 ms using 12.50 MiB
Searched in 221.07 ms using 12.50 MiB
Searched in 220.80 ms using 12.50 MiB
Searched in 223.99 ms using 12.50 MiB
Searched in 223.48 ms using 12.50 MiB

# Search query with DISTINCT
for i in {1..5}; do php bin/bench/search.php 'star wrs'; done

Searched in 84.74 ms using 12.50 MiB
Searched in 83.70 ms using 12.50 MiB
Searched in 87.18 ms using 12.50 MiB
Searched in 84.74 ms using 12.50 MiB
Searched in 85.57 ms using 12.50 MiB
```

> For filter queries (not included above) the performance gap widens for larger result sets, possibly due the cost of cte materialization and temporary b-tree creation

<details>
<summary>sort.php bench script</summary>

```php

<?php

use Loupe\Loupe\SearchParameters;

$config = require_once __DIR__ . '/../config.php';

$options = getopt('d', ['debug']);
$debug = isset($options['d']) || isset($options['debug']);

$startTime = microtime(true);

$searchParameters = SearchParameters::create()
    ->withSort(['release_date:desc'])
;

$result = $config['loupe']->search($searchParameters);

if ($debug) {
    print_r($result->toArray());
}

echo sprintf('Searched in %.2F ms using %.2F MiB', (microtime(true) - $startTime) * 1000, memory_get_peak_usage(true) / 1024 / 1024);
echo PHP_EOL;
```
</details>

**Possible fix**

Would it make sense to only apply `distinct()` when a search query is present? I'm not sure if deduplication is necessary for filter/sort-only queries, but if it isn't, something like this could help:

```php
if (!$tokenCollection->empty()) {
    $qbMatches->distinct();
}
```

Thanks for building Loupe, love using it! Happy to open a PR if this change make sense to you.

ovenum:
Closes #271 

`Searcher::filterDocuments()` applied `->distinct()` unconditionally, even for filter-only or sort-only operations where duplicates should not occur. This caused SQLite to materialize the CTE and build additional temporary indexes unnecessarily.

This change makes `DISTINCT` conditional: it is now applied only when a token collection is present, as duplicate rows should only arise from token-based queries (e.g., full-text matching).

**Benchmarks**
Included are two new benchmarks that can be used to test performance for sort and filter-sort only operations.

| Scenario | Before | After |
|---|---|---|
| Sort only | ~66 ms | ~35 ms |
| Filter Sort | ~50 ms | ~30 ms |

### DIFF
--- a/src/Internal/Search/Searcher.php
+++ b/src/Internal/Search/Searcher.php
@@ -1013,7 +1013,7 @@ class Searcher
     {
         $froms = [];
         $qbMatches = $this->engine->getConnection()->createQueryBuilder();
-        $qbMatches->select('document_id')->distinct();
+        $qbMatches->select('document_id');
 
         // User filters
         $froms[] = $this->filterBuilder->buildFrom();
@@ -1038,6 +1038,13 @@ class Searcher
             $qbMatches->from('(' . $froms[0] . ')');
         } else {
             $qbMatches->from('(' . implode(' INTERSECT ', $froms) . ')');
+        }
+
+        if (!$tokenCollection->empty()) {
+            // Keep DISTINCT here for SQLite performance. Even though the upstream queries already produce one row per
+            // document in practice, making that uniqueness explicit helps SQLite choose a better plan for the final
+            // matches CTE. Removing DISTINCT keeps the results correct, but performs slower.
+            $qbMatches->distinct();
         }
 
         $this->addCTE(new Cte(self::CTE_MATCHES, ['document_id'], $qbMatches));


ovenum:
Sorry for not getting back to this earlier. 
Just ran the new benchmarks included commit 8c26205 against the latest changes on main (commit de3426c). 

Still seeing nearly the same performance differences for `bin/bench/sort.php` and `bin/bench/filter-sort.php`.

| Benchmark | Latest Main | This PR | Latest Main + PR |
|---|---|---|---|
| sort.php | ~66 ms | ~35 ms | ~35ms |
| filter-sort.php | ~50 ms | ~30 ms | ~30ms |
| search.php | ~33ms | ~81ms | ~33ms |

On latest main for full-text queries the improvements are huge though!! Like 50% better performance huge! Congrats for the updates 🎉.  

> The question is more why we need the DISTINCT in the first place 

It might have been added as a performance improvement for full-text search queries. Not applying it when running full-text search will make the query performance worse. But this will not hold true for filter and sort only operations. 

Running `bin/bench/search.php` for latest main

| Benchmark | with DISTINCT | without DISTINCT |
|---|---|---|
| search.php | ~34 ms | ~55 ms |

Sort and filter only benchmarks where not included so this might have gone unnoticed and `distinct` was applied to all queries.

To move ahead with this PR: i could  add tests (and maybe a new fixture) that will ensure no duplicate results are returned when doing full-text search operations that will have matches across multiple fields. But looking at the perf differences, this might have been added as an optimization rather than a guard against duplicate results.

Toflar:
> Sorry for not getting back to this earlier.

Well I'm sorry, I didn't 🙈 



> On latest main for full-text queries the improvements are huge though!! Like 50% better performance huge! Congrats for the updates 🎉.

Awesome! Thanks ❤️ 

> It might have been added as a performance improvement for full-text search queries. Not applying it when running full-text search will make the query performance worse. But this will not hold true for filter and sort only operations.

Indeed, the performance with `DISTINCT` is way better. This is what I would personally add as a comment so we don't forget:

```php
        // Keep DISTINCT here for SQLite performance. Even though the upstream queries already produce one row per
        // document in practice, making that uniqueness explicit helps SQLite choose a better plan for the final
        // matches CTE. Removing DISTINCT keeps the results correct, but performs slower.
```

Apart from that, according to a few tests I think we can also safely remove the `DISTINCT dm.position` in `Relevance.php` as well as the other two (`->distinct()` and `DISTINCT`)  in `Searcher.php`.

> To move ahead with this PR: i could add tests (and maybe a new fixture) that will ensure no duplicate results are returned when doing full-text search operations that will have matches across multiple fields. But looking at the perf differences, this might have been added as an optimization rather than a guard against duplicate results.

Here's the thing: I'm super happy to merge the optimization in this PR and release it. I'm just not happy about the performance benchmarking stuff. Having one or two simple files in `bin/bench` was fine so far but if we keep growing that now, executing them one by one for every PR becomes tedious. So I'd like to switch to a framework like [`phpbench`](https://phpbench.readthedocs.io/en/latest/). However, that's probably going to be a bit of an effort so I won't ask you to do the work but also, I don't have the time to do that at the moment. So I'd suggest you just update the code if you need the fix soonish and we'll see for the phpbench benchmarks later or if you really want to do that major contribution and bring our current benches to a way more future-proof foundation, of course, I won't stop you! 😄

Toflar (approved):

Toflar:
Thanks! I'll release the improvement asap 😊 
Regarding phpbench, I've created a separate issue: https://github.com/loupe-php/loupe/issues/283

PR #277, opened 2026-03-25
Handle nonexistent data directory on startup

daun:
Close #273 and make sure nonexistent data directories are created during setup. Currently, it fails with an exception.

The issue was that `realpath` returns `false` for nonexistent paths. That means an existing path was unintentionally required since a nonexistent path would lead to calling `mkdir` with `false` as its argument. By shuffling things around and calling `realpath` after creating any folders, this is avoided.

Also defines behavior around an empty data dir argument: throw a configuration exception. One option would have been to fall back to `cwd`, but I figure it's better to make this explicit in the consumer code.


Toflar (approved):

Toflar:
GitHub did have another outage so the PR was merged but it's not marked as merged :/
Closing manually - thanks @daun!

PR #282, opened 2026-03-27
fix defaut limit without  pagination

4meck:
why 20?

### DIFF
--- a/src/Internal/Search/AbstractQueryParameters.php
+++ b/src/Internal/Search/AbstractQueryParameters.php
@@ -24,7 +24,7 @@ abstract class AbstractQueryParameters
 
     private ?int $hitsPerPage = null;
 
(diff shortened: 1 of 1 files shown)


4meck:
Think about:
1. return new self(\sprintf('Cannot request more than %d documents per request, use pagination.', SearchParameters::MAX_LIMIT));
maybe always set by default withPagination(1, SearchParameters::MAX_LIMIT);
2. wrong $result->getTotalHits() == SearchParameters::MAX_LIMIT  if limit > SearchParameters::MAX_LIMIT  or limit not set

Toflar:
I'm sorry but I don't understand what you are trying to do.

Toflar:
Closing this because of no feedback and I don't see any reason to adjust the default limit as it's configurable anyway.

PR #284, opened 2026-04-14
Fix highlighting of normalized terms

daun:
- Fixes #274
- Expands on loupe-php/matcher#18
- Store the original `start` and `end` position of terms in documents
  - The naming is deliberate to distinguish it clearly from `position` which is ordinal not positional
- Reuse them during calculation of match spans
- Fixes incorrect highlighting due to term normalization


daun:
Classic, tests only fail in CI :)

daun:
@Toflar Could you pull this branch and run the tests locally? Something is off in CI and I'm losing my mind trying to find out what 🤠 We can probably narrow it down if tests work for you locally at least.

Toflar:
It fails exactly the same locally for me like it does in CI - what do you need me to do? 😊

daun:
@Toflar Great, that's all I needed to know for now. It just means it's me doing something wrong locally. I was assuming it was otherwise, trying to debug CI setup.

daun:
@Toflar One other info might be interesting: what's your local PHP setup for testing? Mac? Homebrew? Nix?

Toflar:
Mac and homebrew, yes :)

daun:
Tried a second machine and it's failing locally as well now. Getting closer 🏓

daun:
I think we're good to go here. Ready for review :)

An additonal fix was required on the matcher side. The discrepancy between local and CI (and local 2) was due to some composer caching shenanigans :/ But I have it resolved now.

Toflar:
Nice, so good call we tried integrating in Loupe first 😉 I'll check the `loupe/matcher` PR first and will let you clean up the `composer.json` stuff in this one after 😊

.github/workflows/ci.yaml
```diff
@@ -99,4 +100,4 @@ jobs:
 
             - name: Output worker.log on failure
               if: ${{ failure() }}
-              run: cat worker.log
\ No newline at end of file
+              run: cat worker.log
```
Toflar: Nitpick, remove this if you can - not so important though

composer.json
```diff
@@ -32,8 +32,14 @@
         "toflar/state-set-index": "^3.0",
         "psr/log": "^2.0 || ^3.0",
         "nitotm/efficient-language-detector": "^3.1",
-        "loupe/matcher": "^0.2.1"
+        "loupe/matcher": "dev-fix/normalization-token-length-on-main"
```
Toflar:
```suggestion
        "loupe/matcher": "^0.2.2"
```

composer.json
```diff
@@ -32,8 +32,14 @@
...
     },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/loupe-php/matcher"
+        }
+    ],
```
Toflar:
```suggestion
```

Toflar (changes requested):
Okay, let's go!

daun:
Done :)

daun:
Ran the benchmarks for fun.

Indexing is 3% slower due to new column inserts (34s → 33s).
Searching is 7% faster, not sure why (129ms → 120ms).

Toflar:
> Searching is 7% faster, not sure why (129ms → 120ms).

With the `bin/bench/search.php`? That should be way lower with the latest optimizations. Let me checkout the PR and double check that.

daun:
Yes. Slow machine, maybe?

daun:
Just checked again, it was overhead by the benchmarking tool (hyperfine). It's 13ms when run directly.

Toflar:
The heck this is fast! 🤣 My machine is slower than yours, both indexing and searching 😊 
Anyway, I couldn't see any performance issue when searching and the additional overhead on the indexing side is to be expected. Giving this a final review now.

Toflar (approved):

Toflar:
Thanks a lot for working on fixing this bug, this was quite some work! ❤️ 
Going to release that to the world now 😊

Issue #286, opened 2026-04-15
Normalization drops relevance info on exactness of spelling


daun:
The recent updates to normalization made it easier to find alternate spellings of terms, but results are no longer correctly ranked based on the exactness of the spelling since we now store normalized terms in the database and a required normalization for matching is not counted as a typo.

**Expected behavior**

Searching for `forderung` should prioritize hits with that exact spelling.
Searching for `förderung` should prioritize that spelling.
The results should be the same, but ranked differently based on the query.

**Observed behavior**

Both searches return results in the exact same order.

**Suggested fix**

I think we should store the original terms in the database, with the normalized spelling added as one variant. That way, the ranking could check if the exact term or a variant was found, and calculate based on that.

I've added a failing test in #287, to be expanded to an actual implementation.

Toflar:
I've already noticed this in January when working on decomposition and had a working POC locally. I just haven't come round to finishing it yet😅

PR #287, opened 2026-04-15
Improve relevance ranking when searching for normalized terms

daun:
This is currently just a failing test for #286.

The goal is to rank documents higher if the pre-normalized spelling matches the exact spelling in the query.

Searching for `müller` should prefer `Thomas Müller` to `Thomas Muller`, and vice-versa.


daun:
Closing in favor of #288

PR #288, opened 2026-04-22
Consider folding state in exactness ranking

### LINKED ISSUE #286: Normalization drops relevance info on exactness of spelling

Toflar:
Attempt to fix https://github.com/loupe-php/loupe/issues/286.
Requires https://github.com/loupe-php/matcher/pull/19

This will have a slight performance impact as the `exactness` ranking algorithm now does more work - which is expected, I guess 😊 

/cc @daun - can you check if that fixes the issue? I've used your tests in this PR but maybe there's other stuff I missed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "toflar/state-set-index": "^3.0",
         "psr/log": "^2.0 || ^3.0",
         "nitotm/efficient-language-detector": "^3.1",
-        "loupe/matcher": "dev-fix/remember-folding-state as 0.2.3"
+        "loupe/matcher": "^0.2.3"
     },
     "require-dev": {
         "phpstan/phpstan": "^2.1.31",
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "toflar/state-set-index": "^3.0",
         "psr/log": "^2.0 || ^3.0",
         "nitotm/efficient-language-detector": "^3.1",
-        "loupe/matcher": "^0.2.2"
+        "loupe/matcher": "dev-fix/remember-folding-state as 0.2.3"
     },
     "require-dev": {
         "phpstan/phpstan": "^2.1.31",
--- a/src/Internal/Engine.php
+++ b/src/Internal/Engine.php
@@ -36,7 +36,7 @@ use Toflar\StateSetIndex\StateSetIndex;
 
 class Engine
 {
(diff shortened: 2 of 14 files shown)


daun:
Looks great! To be super safe, we might test it with another type of folding that changes length, e.g. `ß` → `ss`.

I was assuming this would be less complex to implement 😅 Just thinking out loud, would marking a normalized/folded term as a variant have done the ranking as well? Or is the new `folded` column more exact for this, i.e. there's an advantage to tracking variants vs. folds?

Toflar:
Good call, I added a dedicated regression test for a length-changing fold (`ß -> ss`) to make sure we cover that case as well.
On the second point: marking "variant" terms would not be enough for this ranking problem. Variants are additional terms we index, e.g. stemming-derived alternatives. The exactness issue here is different: it’s about whether the matched term itself had to be folded from the original surface form to match the query. That applies to the main token too, not just to indexed variants.
So the new folded flag is the more precise signal for ranking. It lets us distinguish:

- query/document both unfolded
- query/document both folded
- only one side folded

"Folded" referring to "changed in some way vs. the original ignoring lowercase changes" which is a design decision, I just don't think we should make a difference between e.g. `Thomas` and `thomas` when ranking - doesn't make much sense to consider the case in the query.

That is what exactness needs in order to prefer `Müller` over `Muller` for a `müller` query, and likewise `Straße` over `Strasse` for a `straße` query 😊

daun:
Makes sense! I wasn't thinking about the fold state of both query **and** document tokens. That's useful information.

Toflar:
So you are happy with this bugfix? Then I'll merge this so you get your proper rankings soon 😊

daun:
@Toflar Yup, the fix is exactly what I had in mind!

daun:
Thanks 🎉

PR #289, opened 2026-04-22
Implemented a few indexing performance improvements

Toflar:
Noticed that the index on `position` is useless so we spend time updating this in vain. Also, a little cache around the already seen terms helped a bit too.

Can you verify if that's the case for you too @daun ?


daun:
Yeah, it improves 59s → 49s in indexing, 210ms → 208ms in searching.

daun:
Again, the search values are inflated by hyperfine's overhead.

Toflar:
Good, I have similar results 😎

PR #291, opened 2026-04-22
Fix phrase matching when token appears multiple times

### LINKED ISSUE #290: Phrase search fails in some cases of repeated terms
Phrase group search skips results if the preceding term of a phrase appears somewhere earlier in an attribute.

**Example**

Title: `Gone With the Wind: The Wind Blows No More`

Search: `"the wind"` ✅
Expected: 1 result
Observed: 1 result
Works: both terms in the group are the first appearance in the attribute

Search: `"wind blows"` 🛑
Expected: 1 result
Observed: 0 results
Fails: "wind" already appears once in the attribute before the phrase that should match

daun:
Phrase group search skips results if a term of the phrase already appeared somewhere earlier in an attribute. This changes the logic to consider all possible candidates. Fixes #290.

## Cause

The searcher only considers the first occurrence of a term for checking against valid positions as part of the phrase (`=` comparison). Switching to `IN` and considering all results from the subquery leads to the expected behavior.

```diff
- %s.position = (SELECT position + 1 FROM %s WHERE document=td.document AND attribute=td.attribute)
+ %s.position IN (SELECT position + 1 FROM %s WHERE document=td.document AND attribute=td.attribute)
```

## Example

Title: `Gone With the Wind: The Wind Blows No More`

Search: `"the wind"` ✅
Expected: 1 result
Observed: 1 result
Works: both terms in the group are the first appearance in the attribute

Search: `"wind blows"` 🛑
Expected: 1 result
Observed: 0 results
Fails: "wind" already appears once in the attribute before the phrase that should match

### DIFF
--- a/src/Internal/Search/Searcher.php
+++ b/src/Internal/Search/Searcher.php
@@ -626,7 +626,7 @@ class Searcher
         // Ensure phrase positions if any
         if ($token->isPartOfPhrase() && $previousPhraseToken) {
             $cteSelectQb->andWhere(\sprintf(
-                '%s.position = (SELECT position + 1 FROM %s WHERE document=td.document AND attribute=td.attribute)',
+                '%s.position IN (SELECT position + 1 FROM %s WHERE document=td.document AND attribute=td.attribute)',
                 $termsDocumentsAlias,
                 $this->getCTENameForToken(self::CTE_TERM_DOCUMENT_MATCHES_PREFIX, $previousPhraseToken),
             ));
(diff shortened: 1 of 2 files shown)


Toflar:
Makes sense! Before I merge, any performance issues worth mentioning?

daun:
None from what I can tell. Benchmarks are 2ms faster, but that must be an aberration.

Toflar:
Neat - going to release that now! 😊

PR #292, opened 2026-04-23
Configurable matching strategy: require all terms

daun:
Introduce a new config to allow narrowing search results with each additional term:

`matching_strategy: any | all (default: any)`

## Example

```php
$searchParameters = SearchParameters::create()
    ->withQuery('all of these words must match')
    ->withMatchingStrategy('all');
```

## `any`

Current matching strategy. Only one of the terms must match. Great default for text-heavy applications where users aren't familiar with all of the content: returning more documents with each search term, combined with proper ranking and a threshold, is a good user experience there.

## `all`

New matching strategy: all terms must match. Filters down the result set with each additional term instead of expanding it. Great for projects working with more structured data where editors more or less know what they're looking for, e.g. a database of orders, where searching for a customer name + a city should narrow the results to all orders to that name in that city.

## Prior art

The name of the option is [directly lifted](https://www.meilisearch.com/docs/reference/api/search/search-with-post#matching-strategy) from Meilisearch. There, the default is different: `last` = drop terms from the end of the query until enough results are found. However, `all` is one possible value for the matching strategy.

## Performance

Search performance for multiple terms is identical or slightly faster because less documents are returned in most cases.

Search performance for single terms is unchanged.

### DIFF
--- a/docs/searching.md
+++ b/docs/searching.md
@@ -42,6 +42,21 @@ $searchParameters = \Loupe\Loupe\SearchParameters::create()
 Hint: Note that your query is stripped if it's very long. See the section about [maximum query tokens in the 
 configuration settings][Config].
 
+## Matching strategy
+
+By default, a search will match any documents that match at least one of the query terms. You can change this behavior by configuring a matching strategy.
+
+`any`: A document is returned if it matches at least one of the query terms. This is the default matching strategy.
+
+`all`: A document is only returned if it matches all query terms.
+
+```php
+$searchParameters = \Loupe\Loupe\SearchParameters::create()
+    ->withQuery('all of these words must match')
+    ->withMatchingStrategy(\Loupe\Config\MatchingStrategy::All)
+;
+```
+
 ## Attributes to receive
 
 By default, Loupe returns all the attributes of the documents you've indexed. If you are interested in only a subset 
--- a/docs/searching.md
+++ b/docs/searching.md
@@ -53,7 +53,7 @@ By default, a search will match any documents that match at least one of the que
 ```php
 $searchParameters = \Loupe\Loupe\SearchParameters::create()
     ->withQuery('all of these words must match')
-    ->withMatchingStrategy(\Loupe\Config\MatchingStrategy::All)
+    ->withMatchingStrategy('all')
 ;
 ```
 
--- a/src/Config/MatchingStrategy.php
+++ b/src/Config/MatchingStrategy.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Loupe\Loupe\Config;
+
+enum MatchingStrategy: string
+{
+    case All = 'all';
+    case Any = 'any';
+}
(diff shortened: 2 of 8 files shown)


Toflar:
Cool! Just wondering: this means as a user you cannot influence the strategy yourself (unless there's a UI built for you where you can select but it's not doable via the query itself). However, we already have phrase matching with `"` and negative keywords (excluding) with `-`. Any reason we shouldn't just go for `+`? I remember faintly that this used to be a thing in the early days of Google and others. Just thinking out loud here - not sure it's actually a good idea :D

daun:
I like the `+` idea! Also remember that being a useful thing in pre-2015 Google or so. But I'd prefer having both. The goal here is to influence the default behavior if it makes sense for some projects, regardless of user ability. But telling users to prepend a `+` is a great power user strategy. This PR is (comparatively) small in scope, not sure if the plus sign would mean more or more complex changes?

daun:
I guess the plus sign would require a top-level WHERE for all plus/forced terms, and a nested WHERE with ORs for all the other terms. Not too complex, after all.

Toflar:
> The goal here is to influence the default behavior if it makes sense for some projects, regardless of user ability.

Yes, I do see this use case. Which is exactly why I asked because if you want to force a certain behavior in your project, why would we allow `-` then? 😊 

BTW: This needs to be targeted at `develop` as it's a new feature.

daun:
Interesting point. Personally, I only ever had customers asking for exclusive `AND` queries, not for a `+` modifier.

The `-` modifier is always useful, on the other hand, no? Regardless of whether results are additive or subtractive. Even in `all` mode, leaving out a term would not have the same effect as negating it. That's why I liked the `+` modifier suggestion. It's a different level of control.

Basically, we'd have two levels: the default strategy, and individual modifiers. Depending on whether one adds an AND/OR toggle to the UI, this would give users one or two ways of controlling the strategy.

aPnUb

Screenshot 2026-04-23 at 15 23 52

daun:
Right, I'll switch it to develop!

daun:
Switched to `develop`.

daun:
@Toflar How do you feel about the implementation in general?

I wasn't sure about the new enum in `Loupe\Config`, but also couldn't think of a better place. 

Would you prefer an implementation with `MatchingStrategy` classes, wrapping a Searcher or a query object? The current way of doing it is rather simple, that might be either a strength or a weakness 🤠

Toflar:
> I wasn't sure about the new enum in `Loupe\Config`, but also couldn't think of a better place.

Yeah I don't think it should be placed there. I would place it there if it were a global setting like the typo tolerance - a configuration value for the entire engine. But here, it's part of the search query and we also don't have any constants for `asc` or `desc` etc.

I guess we can move the enum to `Internal/Search` and work with it internally but exposing could be just a string. Even though I actually like using enums very much I'm not sure it fits the current layout well 🤔

daun:
True, the enum feels a bit off. I switched it to expose the matching strategy as string constant.

Toflar (approved):
LGTM, if you're happy with that state, I'll merge as-is 😊

daun:
Good to go from my end :)

Toflar:
Nice! Let's see how long we'll take for 0.14, I hope you can live with `dev-develop` for a bit 😅

daun:
Wonderful. No rush at all with the release!

Toflar:
As discussed, this landed in `main` now and I will release with the next 0.12 release.

PR #295, opened 2026-05-05
Attribute truncation

daun:
- Allow truncating specific attributes
- Make the length (250) and marker (...) adjustable
- One major overhaul was needed: allow returning formatted attributes without also returning their original value. Otherwise, the truncation has no effect on the final amount of data returned.

```php
    ->withAttributesToRetrieve(['id', 'title'])
    ->withAttributesToTruncate(['content']);
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "toflar/state-set-index": "^3.0",
         "psr/log": "^2.0 || ^3.0",
         "nitotm/efficient-language-detector": "^3.1",
-        "loupe/matcher": "^0.2.3"
+        "loupe/matcher": "dev-feature/truncation"
     },
     "require-dev": {
         "phpstan/phpstan": "^2.1.31",
@@ -42,6 +42,12 @@
         "symfony/process": "^7.3.4",
         "symfony/var-dumper": "^7.3.5",
         "symplify/easy-coding-standard": "^12.6.2"
+    },
+    "repositories": {
+        "loupe-matcher": {
+            "type": "vcs",
+            "url": "https://github.com/loupe-php/matcher.git"
+        }
     },
     "scripts": {
         "tests": "@php vendor/bin/phpunit",
--- a/docs/searching.md
+++ b/docs/searching.md
@@ -386,7 +386,7 @@ truncated at the end, so you will get the beginning of the value up to the defin
 
 ```php
 $searchParameters = \Loupe\Loupe\SearchParameters::create()
-    ->withAttributesToTruncate(['overview', 'content']);
+    ->withAttributesToTruncate(['overview']);
 ```
 
 The result of a truncated attribute will look like this:
@@ -406,6 +406,15 @@ $results = [
         ],
     ],
 ];
+```
+
+For very long attributes, it is recommended to remove the attribute from the list of `attributesToRetrieve`
+and only return the truncated version to save bandwidth.
+
+```php
+$searchParameters = \Loupe\Loupe\SearchParameters::create()
+    ->withAttributesToRetrieve(['id', 'title'])
+    ->withAttributesToTruncate(['content']);
 ```
 
 The default truncation length is 250 characters. You can define a different truncation length by passing in an
--- a/docs/searching.md
+++ b/docs/searching.md
@@ -379,9 +379,60 @@ $results = [
 ];
 ```
 
+## Attribute truncation
+
+Loupe can truncate selected attributes to a certain length to avoid returning excessively long values. Attributes are
+truncated at the end, so you will get the beginning of the value up to the defined truncation length.
+
+```php
+$searchParameters = \Loupe\Loupe\SearchParameters::create()
+    ->withAttributesToTruncate(['overview', 'content']);
+```
+
+The result of a truncated attribute will look like this:
+
+```php
+$results = [
+    'hits' => [
+        [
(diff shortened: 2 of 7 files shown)


Toflar:
Looks really cool!

First of all, I think we're going to make an organisational change here. There's no point in releasing a 0.13 if there are non-breaking changes in 0.12 - it forces people to update their dependency definitions for no reason. It's like releasing a new major version with no breaks. So I guess with the next 0.12 version, we'll switch from only releasing bugfixes to also releasing features until we finally hit v1.0 where we'll do proper semantic versioning. So you can target `main` here, I'll make sure your other stuff that's already in `develop` will land in `main` asap too.

Regarding this PR:

Implementation quality looks fantastic! ❤️ Just so we get the naming and config right. I tried searching if `truncation` is what other search engines use and I couldn't find anything like that. What I could find though is that ElasticSearch apparently also let's you configure the individual match cropping length (called `fragment_size`, defaults to `100`) and then they don't let you specify the truncation length but instead the number of "fragments" (called `number_of_fragments`, defaults to `5`) (https://www.elastic.co/docs/reference/elasticsearch/rest-apis/highlighting-settings). So you don't specify two lengths but instead just the crop length and then the desired max. of fragments used - maybe we should do that instead? Wdyt?


> * One major overhaul was needed: allow returning formatted attributes without also returning their original value. Otherwise, the truncation has no effect on the final amount of data returned.

Is this really needed? We already have `attributesToCrop` and if I'm not mistaken, this is exactly what Meilisearch would use too?

```
{
  "attributesToRetrieve": ["id", "title"],
  "attributesToCrop": ["content"],
  "cropLength": 20
}
```

daun:
> First of all, I think we're going to make an organisational change here. There's no point in releasing a 0.13 if there are non-breaking changes in 0.12

Sounds good!

> So you don't specify two lengths but instead just the crop length and then the desired max. of fragments used - maybe we should do that instead? Wdyt?

That's a nice mental model. How does it handle attributes that don't have any of the search terms? I guess they just won't be in the formatted array then? That might also answer the other question: 

> Is this really needed? We already have attributesToCrop and if I'm not mistaken, this is exactly what Meilisearch would use too?

The way I understood and implemented it is that the truncation should happen on the `_formatted` attribute. Are you saying that the truncation should rather apply to the original document attribute, no matter how many matches are found in it?

daun:
If we went this route, it could be called something like `->withMaxCropFragments(5)` or so.

daun:
Mmh, does the max count really solve the original problem, though?

If I want to show a summary of a document's content in search results, I need to consider a scenario where the only matches are in the title. Even then, I still want to show a content summary. So I need to return the document's `content` key, but it might be too big to send in the request. That's why I'm truncating it.

If we do this by adding a max count to the cropping feature, wouldn't that mean it won't apply to the `content` attribute because there are no matches?

Toflar:
Maybe there's a misunderstanding here but `cropLength` as well as `cropCount` are per attribute. 

> The way I understood and implemented it is that the truncation should happen on the `_formatted` attribute. Are you saying that the truncation should rather apply to the original document attribute, no matter how many matches are found in it?

No no, to the formatted but afaics we don't need a new `withAttributesToTruncate()` setting - we can re-use the already existing `withAttributesToCrop()`.

daun:
> No no, to the formatted but afaics we don't need a new withAttributesToTruncate() setting - we can re-use the already existing withAttributesToCrop().

So the cropping happens around matches, right? If there are no matches in the `content` attribute, it won't be in `_formatted`. How would I display a truncated version of it anyway for my results list? That's why I figured we might need to truncate the original attribute instead. Truncating doesn't depend on matches; cropping does, though. Or maybe you're saying that cropping should fall back to truncating in case of no matches?

Toflar:
> So the cropping happens around matches, right? If there are no matches in the `content` attribute, it won't be in `_formatted`. How would I display a truncated version of it anyway for my results list?

So you want to show a truncated value of an attribute you've indexed with no matches? Why would you only truncate that on output and not already on indexing time? It would cause e.g. a string of 200kb to be read by loupe only to then get truncated on every request? Should probably have a `content_truncated` in that case and truncate when indexing, no?

Toflar:
I think this would be better in combination with `displayedAttributes` anyway. So you don't add your `content` to `displayedAttributes` which means it doesn't get added to the document data in the `documents` table and thus Loupe doesn't need to load it from sqlite when returning the results. 
So you'd have `->withSearchableAttributes(['content'])->withDisplayedAttributes(['content_truncated'])` and during index time you provid both, `['content' => '200 kb string', 'content_truncated' => '250 chars string']`.
Then, when searching you would output `$result['_formatted']['content'] ?? $result['content_truncated']`.

daun:
> Then, when searching you would output $result['_formatted']['content'] ?? $result['content_truncated'].

That's a good strategy, didn't think of that. And it would get rid of the need for a pure truncation setting. Limiting the cropping to a maximum number of fragments is all we need. Nice :)

Toflar:
I am very sorry I only noticed this now after all your work - sometimes my brain needs code to find the best solution, I guess... :/

Toflar:
I think I'll create a PR with that strategy for the performance part of the docs 😊

daun:
Closing this in favor of a (future) `->withCropMaxFragments()` settings.

PR #296, opened 2026-05-06
Added support for a query cache with APCu support out-of-the-box

Toflar:
This PR adds query-time state-set cache reuse for typo-tolerant search, with a zero-config acceleration when APCu is available.

I intentionally called it `query cache` so should we have other cacheable information in the future, we can add that to the same cache but for now, it's basically "only" caching state set caculation. It depends on https://github.com/Toflar/state-set-index/pull/21 which contains the feature to calculate a state snapshot and calculate the states from a given snaphsot.

Why do I want to do that?

Autosubmit/typeahead queries are highly similar (`thoma` -> `thomas`, `amakin dkywalke` -> `amakin dkywalker`).  
Without reuse, state-set matching is recalculated for each request. This is especially wasteful for multi-term queries where usually only the last term changes vs. a prior request.
This PR reuses prior prefix state snapshots within a short TTL window of 60 seconds.

Changes summarized:

- Introduced a query cache in configuration via PSR-6
- Added default internal APCu PSR-6 pool (auto-enabled when APCu is available, disable with `withQueryCache(null)`)
- Added index-UID namespacing to avoid cross-index cache key collisions
- Added versioned cache invalidation on `index()` / `removeFromIndex()`
- Added bounded prefix reuse for typeahead (`-1`, `-2`) and only for the last query term. This means that when searching for `thomas` we can re-use a previous `thoma` cache in case this was already calculated.

Performance is a bit hard to measure and very much depends on the query (longer terms = more typos are tolerated = more states need to be calculated = caching helps a lot more).
I tried with our `php bin/bench/search.php` example (so the `Amakin Dkywalke` vs  `Amakin Dkywalker`) tests in a web environment (because on CLI there's no APCu request cache sharing) and was able to get about **35% to 40%** faster search results as of the second search request 🚀 😎

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,8 @@
         "wamania/php-stemmer": "^4.0",
         "doctrine/lexer": "^2.0 || ^3.0",
         "mjaschen/phpgeo": "^5.0 || ^6.0",
-        "toflar/state-set-index": "^3.0",
+        "toflar/state-set-index": "^3.3",
+        "psr/cache": "^2.0 || ^3.0",
         "psr/log": "^2.0 || ^3.0",
         "nitotm/efficient-language-detector": "^3.1",
         "loupe/matcher": "^0.2.3"
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -216,6 +216,30 @@ $configuration = \Loupe\Loupe\Configuration::create()
 
 If you need to browse all the documents and ignore this limit, use [the browse API][Browse].
 
+## Query cache
+
+Loupe can cache internal query artifacts (such as state-set lookups during typo tolerant search) through
+a PSR-6 cache pool:
+
+```php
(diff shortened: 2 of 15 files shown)


Toflar:
This is ready and could be merged imho.
/cc @daun - wdyt? 😊

daun:
Nice! Can't try it out currently, but implementation looks great 👌

Could one re-use an existing Symfony cache adapter from Symfony also for the APCu cache, same as the Redis example? Or is there a benefit to creating our own `ApcuCachePool`? I assume we just don't have the package installed at the moment, but merely curious.

Toflar:
Yeah essentially I couldn't find any APCu adapter library with basically 0 dependencies except for PSR. There are a few but `symfony/cache` is certainly not one of those - it ships with a ton of adapters which I don't think is a good fit for us. Also it was pretty easy to implement so...can still switch to a library if we want to.

PR #297, opened 2026-05-06
Document preview-attribute fallback for unmatched cropped content

Toflar:
As promised in https://github.com/loupe-php/loupe/pull/295 /cc @daun


daun:
🎉

Issue #283, opened 2026-04-07
Use phpbench for benchmarks


Toflar:
> I'm just not happy about the performance benchmarking stuff. Having one or two simple files in `bin/bench` was fine so far but if we keep growing that now, executing them one by one for every PR becomes tedious. So I'd like to switch to a framework like [`phpbench`](https://phpbench.readthedocs.io/en/latest/). However, that's probably going to be a bit of an effort so I won't ask you to do the work but also, I don't have the time to do that at the moment. So I'd suggest you just update the code if you need the fix soonish and we'll see for the phpbench benchmarks later or if you really want to do that major contribution and bring our current benches to a way more future-proof foundation, of course, I won't stop you! 😄

_Originally posted by @Toflar in https://github.com/loupe-php/loupe/issues/272#issuecomment-4164024912_

ovenum:
Not familiar with `phpbench` and had a look at it out of curiosity. 

If you've already got a good picture of what should be benched and how let me know. For now I've replicated the base scenarios from  `bin/bench` with the movies.json dataset and added sort and sort-filter scenarios.

It's still _eye  ball_ only so no GitHub actions integration or regression tests. 

It could also benefit from a quick way to compare performance of branch a to branch b. 

Have to clean this up and should have a draft PR by end of week with basic functionality.

Toflar:
> If you've already got a good picture of what should be benched and how let me know. For now I've replicated the base scenarios from `bin/bench` with the movies.json dataset and added sort and sort-filter scenarios.

Something along the lines of https://github.com/Toflar/state-set-index/pull/18/. So yes, I would've converted the existing benchs in a first PR and then extended with yours in a second one.
And no automation via GitHub Actions yet :)

Toflar:
Implemented in https://github.com/loupe-php/loupe/pull/300 🤩

PR #300, opened 2026-05-28
PHPBench

daun:
- Set up PHPBench for Loupe
- Transfers existing benchmarks
- Separate query benchmarks (`composer bench`) from slow index benchmarks (`composer bench-index`)
- Splits indexing benchmark into 1k and 10k documents
- All 40k documents would take ~50s and don't lend themselves to a quick performance check

### Query

  | Benchmark | Subject | Mode | rstdev | mem_peak |
  | --- | --- | --- | --- | --- |
  | `SearchBench` | `benchPlainQuery` | 208.32 ms | ±2.02% | 3.12 MB |
  | `SearchBench` | `benchTypoQueryWithFacets` | 34.67 ms | ±6.80% | 7.81 MB |
  | `SearchBench` | `benchExactQueryWithFacets` | 33.20 ms | ±2.59% | 7.81 MB |
  | `FilterBench` | `benchFilteredSortedSearch` | 36.67 ms | ±1.36% | 7.88 MB |

### Index

  | Benchmark | Subject | Set | Revs | Its | Mode | rstdev |
  | --- | --- | --- | --- | --- | --- | --- |
  | `IndexBench` | `benchAddDocuments` | 1k | 1 | 1 | 414.62 ms | — |
  | `IndexBench` | `benchAddDocuments` | 10k | 1 | 1 | 2,986.23 ms | — |
  | `IndexBench` | `benchUpdateDocuments` | 1k | 1 | 1 | 399.48 ms | — |
  | `IndexBench` | `benchUpdateDocuments` | 10k | 1 | 1 | 3,050.65 ms | — |


tests/Benchmark/FilterBench.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Loupe\Loupe\Tests\Benchmark;
+
+use Loupe\Loupe\Loupe;
+use Loupe\Loupe\SearchParameters;
+
+/**
```
Toflar: Those should be all converted to php attributes.

tests/Benchmark/IndexBench.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Loupe\Loupe\Tests\Benchmark;
+
+use Loupe\Loupe\Loupe;
+
+/**
+ * @BeforeClassMethods({"setUpClass"})
```
Toflar: Same here :)

tests/Benchmark/IndexBench.php
```diff
@@ -0,0 +1,87 @@
...
+
+    public function provideCorpus(): \Generator
+    {
+        yield '1k' => [
+            'size' => 1_000,
+        ];
+        yield '10k' => [
+            'size' => 10_000,
+        ];
+        // yield 'all' => ['size' => 0];
```
Toflar: Remove? Or did you leave that here so we can run it on demand? Sounds like there should be some kind of grouping that we can then run on CLI? Like in phpunit with `--testsuite=very-slow`?

tests/Benchmark/SearchBench.php
```diff
@@ -0,0 +1,100 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Loupe\Loupe\Tests\Benchmark;
+
+use Loupe\Loupe\Loupe;
+use Loupe\Loupe\SearchParameters;
+
+/**
```
Toflar: PHP attributes.

tests/Benchmark/SearchBench.php
```diff
@@ -0,0 +1,100 @@
...
+
+    public function benchPlainQuery(): void
+    {
+        $this->loupe->search(
+            SearchParameters::create()->withQuery('star wars')
+        );
+    }
+
+    /**
+     * @Revs(1)
```
Toflar: Attribute.

phpbench.json
```diff
@@ -0,0 +1,12 @@
+{
+    "$schema": "vendor/phpbench/phpbench/phpbench.schema.json",
+    "runner.bootstrap": "vendor/autoload.php",
+    "runner.path": "tests/Benchmark",
+    "runner.file_pattern": "*Bench.php",
+    "runner.php_config": {
```
Toflar, line 6: Care to explain why you replaced the defaults here?

README.md
```diff
@@ -39,10 +39,20 @@ Performance depends on many factors but here are some ballpark numbers based on
 * **Indexing** will take around **60 seconds** (this varies greatly because it depends on how much content per document
   you want to index
 
-Note that anything above 50k documents is probably not a use case for Loupe. You can run your own benchmarks 
-using the scripts in the `bin/bench` folder: `index.php` for indexing and `search.php` for searching. 
-Please, also read the [Performance](./docs/performance.md) chapter in the docs. You may report your own performance 
-measurements and more details in the [respective discussion][Performance_Topic]. 
+Note that anything above 50k documents is probably not a use case for Loupe. You can run your own benchmarks
```
Toflar, line 42: Maybe we should generalize that statement anyway - I have indexes with 100k documents and Loupe just doesn't care at all 😄 The more documents, the slower it will be with default configuration but I guess it also depends on the size of those documents of course etc. - care to give that a shot?

Toflar (commented):
Nice! ❤️

tests/Benchmark/FilterBench.php
daun: Done :)

tests/Benchmark/SearchBench.php
daun: Done :)

tests/Benchmark/SearchBench.php
daun: Done :)

tests/Benchmark/IndexBench.php
daun: I meant to reenable that actually. I've already split the benchmarks into two groups: query (default) and index. They are run using `composer run bench` for the default query group and `composer run bench-index` for the slow index group. So safe to re-enable that corpus line.

phpbench.json
daun: Yes, since we test in PHP 8.3 and 8.4 and the OPCache defaults changed in 8.4, i wanted to actively avoid surprises should we decide to run the benchmarks in GitHub actions at some point. And the memory_limit was necessary on larger index tests, at least on my machine. We might be able to lower that again, though.

README.md
daun: Good idea! I'd say we can finish the benchmarks and then test out where Loupe's limits are with a bigger dataset. Or did you mean to just update the readme with a note? Either way, I'd say that's a different PR maybe.

tests/Benchmark/IndexBench.php
daun: Reenabled it.

phpbench.json
Toflar: The memory looks weird. I mean indexing should not take up that much memory - is this due to overhead of phpbench or Loupe itself?

daun: Must be phpbench overhead since Loupe usually indexes the whole dataset in less than 100MB.

daun: Checked again, the default crashes but 128M seems to be enough now. Guess I just wanted to make sure :) Lowering to 128M.

daun: Spoke too early. 128M crashes too. Checking 256M now.

daun: Or maybe a memory leak when indexing the whole database multiple times?

daun: 256M seems enough for all indexing tests. I lowered it. We can/should check if that's a memory leak when indexing/reindexing multiple times in a row, but for now we know they will all go through with that setup :)

daun:
Indexing all documents consumes 103M of memory. Indexing + reindexing consumes 120M. That plus overhead of PHPBench probably leads to the 128M memory not being enough to run the indexing benchmarks.

Toflar (approved):

Toflar:
Super cool, we can finally close https://github.com/loupe-php/loupe/issues/283 now! Thanks, Philipp!

PR #301, opened 2026-05-30
Use shared candidate document set

daun:
- Compute a shared `_cte_candidate_documents` CTE across all tokens
- The matches-any-term set is now materialized & indexed once instead of being derived in every token match CTE
- Speedup of around 35% for multi-term & 3% for single-word queries

### Why

- The candidate-document set is the union, over all tokens, of each token's `_cte_term_documents_<token>` row
- Currently, that set is built inline in every per-token match CTE as a subquery
- With this change, the union is built once as a shared CTE, and each per-token match CTE references it a single time
- It's the same logical filter, but computed once over a materialized indexable set and checked once per token
- Bonus: we can now short-circuit positive single-token group searches to use that new CTE

### Benchmark

| Subject | Before | After | Change |
|---|---|---|---|
| benchExactQueryWithFacets | 33,71ms | 23,15ms | **31% faster** |
| benchMultiWordQueries | 1.602,61ms | 994,57ms | **38% faster** |
| benchPlainQuery | 209,66ms | 123,34ms | **41% faster** |
| benchSingleWordQueries | 282,32ms | 273,06ms | **3% faster** |
| benchTypoQueryWithFacets | 32,03ms | 23,39ms | **27% faster** |
| benchFilteredSortedSearch | 31,21ms | 31,32ms | **0,4% slower** (margin of error) |

### Notes

I've added an explanation of the current CTE pipeline at the top because I kept forgetting which CTE does what :)

### DIFF
--- a/src/Internal/Search/Searcher.php
+++ b/src/Internal/Search/Searcher.php
@@ -33,6 +33,8 @@ class Searcher
 {
     public const CTE_ALL_MULTI_FILTERS_PREFIX = '_cte_mf_all_';
 
+    public const CTE_CANDIDATE_DOCUMENTS = '_cte_candidate_documents';
+
     public const CTE_MATCHES = '_cte_matches';
 
     public const CTE_TERM_DOCUMENT_MATCHES_PREFIX = '_cte_term_document_matches_';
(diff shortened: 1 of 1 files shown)


Toflar (approved):

Toflar:
So cool! When I first worked on Loupe I was happy with anything below 500ms and look where we landed at now!🤩🤯

PR #302, opened 2026-05-31
Make benchmarks more stable

Toflar:
JIT is incompatible with extensions like `xdebug` - see https://bugs.xdebug.org/view.php?id=2293.
I also think it's better for benchmarks anyway because it keeps them more stable.

Okay for you @daun?


daun:
Sounds good. Are you seeing any changes in the benchmark numbers with that?

Toflar:
No, if anything they are just more constant.

PR #305, opened 2026-05-31
Calculate exactness data once per term

daun:
Another quick win for performance: calculate typo/exactness data once per term.
Seeing a speedup of 50-75% for multi-word queries, and 12% for the typo query.
Loving the benchmarks :)

### The problem

- Currently, levenshtein distance is calculated once per occurrence (document x term)
- A common term appearing in thousands of documents had its distance computed thousands of times

### What this does

- Do the expensive calculation **once per matching term** instead of **once per matched occurrence**
- Move the data into the term-matches CTE and carry it forward for later reuse
- That CTE can be joined more cheaply than the terms-documents table, allowing primary key usage on the inner join

### Benchmarks

| Benchmark   | Subject                   | Before Mode | After Mode | Change  |
|-------------|---------------------------|-------------|------------|---------|
| SearchBench | benchExactQueryWithFacets | 9.56ms      | 8.72ms     | -8.79%  |
| SearchBench | benchMultiWordQueries     | 629.95ms    | 293.16ms   | -53.46% |
| SearchBench | benchPlainQuery           | 100.94ms    | 22.73ms    | -77.48% |
| SearchBench | benchSingleWordQueries    | 141.18ms    | 76.59ms    | -45.75% |
| SearchBench | benchTypoQueryWithFacets  | 9.06ms      | 8.56ms     | -5.52%  |
| FilterBench | benchFilteredSortedSearch | 19.49ms     | 19.83ms    | +1.74%  |

### DIFF
--- a/src/Internal/Search/Searcher.php
+++ b/src/Internal/Search/Searcher.php
@@ -614,36 +614,28 @@ class Searcher
         $termsAlias = $this->engine->getIndexInfo()->getAliasForTable(IndexInfo::TABLE_NAME_TERMS);
 
         if ($needsTypoCount) {
-            $cteSelectQb->addSelect(\sprintf(
-                'MIN(loupe_levensthein(%s.term, %s, %s)) AS typos',
-                $termsAlias,
-                $this->createNamedParameter($token->getTerm()),
-                $this->engine->getConfiguration()->getTypoTolerance()->firstCharTypoCountsDouble() ? 'true' : 'false'
-            ));
+            $cteSelectQb->addSelect(\sprintf('MIN(%s.typos) AS typos', $termMatchesCTE));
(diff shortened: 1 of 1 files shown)


daun:
Trying to merge the recent changes from #304, really tricky since they touched the same things 🤠

Toflar:
> Trying to merge the recent changes from #304, really tricky since they touched the same things 🤠

Sorry 🙈

daun:
This PR actually got simpler by your recent changes 🤠 Force-pushed a new version that maps onto main and updated the benchmark values above.

Toflar (approved):

Toflar:
Nice one!! I think I also found another one 😊

PR #311, opened 2026-05-31
Skip redundant candidate-doc filter for single-token queries

Toflar:
Optimize single-token query execution by avoiding the `_cte_candidate_documents` membership filter in `addTermDocumentMatchesCTE()` when only one search token is present.
For single-token searches, this filter is redundant and adds unnecessary query work. We now keep it only for multi-token queries.
Benchmark impact:
* benchSingleWordQueries: ~51.5ms -> ~34.7ms
* benchMultiWordQueries: effectively unchanged

/cc @daun


daun:
🥂

PR #316, opened 2026-06-01
Fix phrase group search

daun:
Somewhere along the recent performance improvements, we slowed down "phrase group" searches catastrophically. They are fine in the latest release, but time out in the current state of the main branch.

This switches things around to use a cross join for checking phrase conditions. What's nice is we can skip the candidate-document query for all following tokens of a group since the first token already ensures the document, we just need to check the position.

Added phrase group benchmarks. We didn't have any, so that's why we missed the performance regression.

### Benchmarks

0.13.17: 1.617ms
main: 30s (timed out)
this: 35ms

### DIFF
--- a/src/Internal/Search/Searcher.php
+++ b/src/Internal/Search/Searcher.php
@@ -1471,10 +1471,16 @@ class Searcher
         if ($this->ctesByName !== []) {
             $queryParts[] = 'WITH';
             foreach ($this->ctesByName as $name => $cte) {
+                $materializedHint = match ($cte->isMaterialized()) {
+                    true => 'MATERIALIZED ',
+                    false => 'NOT MATERIALIZED ',
+                    null => '',
+                };
                 $queryParts[] = \sprintf(
-                    '%s (%s) AS (%s)',
+                    '%s (%s) AS %s(%s)',
                     $name,
                     implode(',', $cte->getColumnAliasList()),
+                    $materializedHint,
                     $cte->getQueryBuilder()->getSQL()
                 );
                 $queryParts[] = ',';
--- a/src/Internal/Search/Searcher.php
+++ b/src/Internal/Search/Searcher.php
@@ -635,7 +635,7 @@ class Searcher
 
         $isPhraseContinuation = $token->isPartOfPhrase() && $previousPhraseToken !== null;
         if ($isPhraseContinuation) {
-            // Continuuing inside "a phrase": drive query from small materialized previous token's match CTE
+            // Continuing inside "a phrase": drive query from small materialized previous token's match CTE
             // Use CROSS JOIN instead of INNER JOIN to pin join order and avoid full scans for common words like "his"
             $previousPhraseCte = $this->getCTENameForToken(self::CTE_TERM_DOCUMENT_MATCHES_PREFIX, $previousPhraseToken);
             $previousPhraseAlias = $previousPhraseCte . '_prev';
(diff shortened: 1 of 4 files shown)


src/Internal/Search/Searcher.php
```diff
@@ -633,18 +633,41 @@ private function addTermDocumentMatchesCTE(Token $token, ?Token $previousPhraseT
...
-        $cteSelectQb->from($termMatchesCTE);
-        $cteSelectQb->innerJoin(
-            $termMatchesCTE,
-            IndexInfo::TABLE_NAME_TERMS_DOCUMENTS,
-            $termsDocumentsAlias . ' INDEXED BY sqlite_autoindex_terms_documents_1',
-            \sprintf('%s.id = %s.term', $termMatchesCTE, $termsDocumentsAlias)
-        );
+        $isPhraseContinuation = $token->isPartOfPhrase() && $previousPhraseToken !== null;
+        if ($isPhraseContinuation) {
+            // Continuuing inside "a phrase": drive query from small materialized previous token's match CTE
```
Toflar:
```suggestion
            // Continuing inside "a phrase": drive query from small materialized previous token's match CTE
```

tests/Benchmark/SearchBench.php
```diff
@@ -82,7 +112,6 @@ public function benchSingleWordQueries(): void
         foreach ($queries as $q) {
             $this->loupe->search(SearchParameters::create()->withQuery($q));
         }
-
```
Toflar, line 85: CS?

src/Internal/Search/Searcher.php
```diff
@@ -633,18 +633,41 @@ private function addTermDocumentMatchesCTE(Token $token, ?Token $previousPhraseT
...
+            // Use CROSS JOIN instead of INNER JOIN to pin join order and avoid full scans for common words like "his"
+            $previousPhraseCte = $this->getCTENameForToken(self::CTE_TERM_DOCUMENT_MATCHES_PREFIX, $previousPhraseToken);
+            $previousPhraseAlias = $previousPhraseCte . '_prev';
+            $cteSelectQb->from(\sprintf(
+                '%1$s %2$s'
```
Toflar: Can we not use positional placeholders? We don't use them anywhere so far and I think it makes it very hard to read.

Toflar (changes requested):

tests/Benchmark/SearchBench.php
daun: CS didn't catch the empty line, so I removed it manually. Maybe we need another setting to fix those too.

daun: I've added a `NoExtraBlankLinesFixer` ruleset to ECS.

daun: I've removed it again because it revealed different settings for local vs CI I think. Will look at this in another PR.

src/Internal/Search/Searcher.php
daun: Switched to using repeated arguments.

Toflar (approved):

Toflar:
Nice thing we have benchmarks now! Helps a lot with LLMs :)

daun:
@Toflar Interestingly, I haven't had much luck getting Claude to use existing benchmarks. Whenever i ask it to profile something, it usually insists on writing its own set of benchmarks from scratch 😅 Hopefully it at least takes the existing ones as inspiration...

PR #317, opened 2026-06-03
Prioritize and limit crop windows

daun:
Add two related features:

- Prioritize matches ("smart cropping"): let the formatter rank and choose the best crop windows
- Max crop fragments: limit number of crop windows to return in formatted results

Update (Jul 30)

- Match priorization enabled by default, the flag must now be toggled off to enforce strict document order
- Max crop fragments set to 5 by default

### Prioritize matches

Cropping will now rank areas by the number of distinct keywords and render them in document order.

```php
    ->withAttributesToCrop(['title', 'summary'], prioritizeMatches: true);
```

### Max crop fragments

Return no more than `{max}` crop windows in formatted results

```php
    ->withAttributesToCrop(['title', 'summary'], cropMaxFragments: 3);
```

## Example for priorization

Given one document, searching for `alpha bravo charlie` with `cropMaxFragments: 2`:

> In the opening the term **alpha** appears all by itself. […] Midway through the document the terms **alpha** **bravo** appear side by side. […] Near the very end **alpha** **bravo** **charlie** cluster tightly together.

The three match windows have increasing density: 1 term → 2 terms → 3 terms

### Default behavior

Fragments are picked in document order*. The two earliest win, the densest one is dropped:

> …opening the term **alpha** appears all by…the terms **alpha** **bravo** appear side…

### Prioritizing matches

The two densest windows are selected (2 terms, then 3 terms) and they're still emitted in document order, so the first fragment is dropped and the best fragment comes last.

> …**alpha** **bravo** appear side by side.…end **alpha** **bravo** **charlie** cluster…

## Base branch

~~Needed to base this on the decomposition branch to get a clean diff. The decomposition stuff is already part of `loupe-matcher#main` and can no longer be cleanly separated. We need to merge the decomposition feature first so this works, but that's doable :)~~


Toflar:
Hello @daun, sorry it took me so long but we're here now - everything you need should be in `main` so you can finish this PR whenever you feel like 😎

daun:
@Toflar Alright, I'll look into resolving the merge conflicts.

Toflar:
Also, I think we might want to enable prioritizing by default - or any reason not to?

daun:
@Toflar Prioritization only has an effect if we're also limiting the windows since they're emitted in document order. Without a limit, it's basically all of them. So we'd need to define a default crop window limit. I think that'd make sense anyway, but we should probably benchmark the window selection logic to see if enabling it by default negates the benefits of less data returned.

daun:
@Toflar I've merged the recent changes from the main branch. Ready for review :)

tests/Functional/SearchTest.php
```diff
@@ -3002,6 +3002,7 @@ public function testRelevanceAndRankingScore(): void
     public function testRelevanceAndRankingScoreForLengthChangingFolding(): void
     {
         $configuration = Configuration::create()
+            ->withLanguages(['de'])
```
Toflar: Why those changes?

tests/Functional/SearchTest.php
```diff
@@ -3137,6 +3138,7 @@ public function testRelevanceAndRankingScoreForNonExistentQueryTerms(): void
     public function testRelevanceAndRankingScoreForNormalizedSpelling(): void
     {
         $configuration = Configuration::create()
+            ->withLanguages(['de'])
```
Toflar: Same here?

tests/Fixtures/Data/movies_1000_de.json
```diff
@@ -0,0 +1,1001 @@
+[
+  {"id":2,"title":"Ariel","overview":"Taisto Kasurinen ist ein finnischer Bergarbeiter, dessen Vater gerade Selbstmord begangen hat und der f\u00fcr ein Verbrechen verantwortlich gemacht wird, das er nicht begangen hat. Im Gef\u00e4ngnis beginnt er davon zu tr\u00e4umen, das Land zu verlassen und ein neues Leben zu beginnen. Er flieht aus dem Gef\u00e4ngnis, aber die Dinge laufen nicht wie geplant...","genres":["Drama","Crime","Comedy"],"poster":"https://image.tmdb.org/t/p/w500/ojDg0PGvs6R9xYFodRct2kdI6wC.jpg","release_date":593395200},
```
Toflar: Why this file?

Toflar (changes requested):

tests/Functional/SearchTest.php
daun: I think that's a leftover from when the PR was based on the decompounding branch. It was necessary to make sure the compounding worked correctly even when Loop wasn't confident about the language of the query itself. I'll try removing it and see if it fails.

tests/Functional/SearchTest.php
daun: See note above.

tests/Fixtures/Data/movies_1000_de.json
daun: Oh, that was supposed to be gitignored, only used for faster testing. I'll remove it :)

tests/Functional/SearchTest.php
daun: [This is the commit](https://github.com/loupe-php/loupe/commit/3772a8888c11eb45a451ca3f1e51f37ab661a6ae) where you added those two lines. Maybe they have been removed since, and the merge made them reappear 🤠

Toflar: Yes in https://github.com/loupe-php/loupe/pull/320.

Toflar:
LGTM now! The only thing remaining is the "should we enable this by default" question:

> So we'd need to define a default crop window limit. I think that'd make sense anyway, but we should probably benchmark the window selection logic to see if enabling it by default negates the benefits of less data returned.

Did you run some benches?

daun:
@Toflar Not yet. Will do now :)

daun:
@toflar Okay, been doing some benchmarks on my local copy.

I'm seeing almost no difference through prioritization; it's around plus or minus 1%.

I also just noticed that the current implementation of this branch defaults to a max crop fragments value of 5 anyway. Since it doesn't make any noticeable difference performance-wise, we should definitely enable prioritization by default.

What's your opinion? Still in favor of enabling it by default as well?

Toflar:
> What's your opinion? Still in favor of enabling it by default as well?

Totally, it means better highlighting context for everyone - why wouldn't we want to have that by default if it comes basically for free? 😊

daun:
@Toflar Nice :) I've switched it to on by default and updated the README for some clarification. I also added a feature note to the base README, since I think it's quite a nice feature to advertise. Ready for review.

Toflar (approved):

Toflar:
Faaaancy! ❤️

PR #326, opened 2026-07-09
Add fallback for SQLite < 3.35.0 without RETURNING support

### LINKED ISSUE #325: loupe 0.13 requires SQLite ≥ 3.35 (RETURNING) with no fallback — breaks on shared hosting (IONOS: PHP 8.4, SQLite 3.34.1)
### Summary

Since **v0.13**, `BulkUpserter::execute()` unconditionally takes the "modern" upsert path (`executeModern()`), which emits an SQL **`RETURNING`** clause. `RETURNING` requires **SQLite ≥ 3.35.0** (released 2021-03-12). There is no fallback for older SQLite — only a comment where one was anticipated, right in `execute()`:

```php
// Modern path: INSERT .. ON CONFLICT .. DO UPDATE [RETURNING]
$results = [...$results, ...$this->executeModern($chunk, $updateColumns)];

// Depending on the version, we could introduce fallback solutions here
```

On hosts whose PHP ships an older bundled SQLite (very common on shared hosting), **every index write throws**, which makes Loupe unusable there. v0.12 worked fine on the exact same host.

### Environment (where it breaks)

- Host: **IONOS** shared hosting
- PHP: **8.4.22**
- PDO_SQLite / SQLite: **3.34.1**  ← one release short of the 3.35.0 that `RETURNING` needs
- Loupe: **0.13.21** (used via `arnoson/kirby-loupe` 0.4.1 on Kirby CMS)

### Error

```
PDOException: SQLSTATE[HY000]: General error: 1 near "RETURNING": syntax error
```

Abbreviated stack:

```
Loupe\Loupe\Internal\Index\BulkUpserter\BulkUpserter->executeQuery()   BulkUpserter.php
Loupe\Loupe\Internal\Index\BulkUpserter\BulkUpserter->executeModern()  BulkUpserter.php
Loupe\Loupe\Internal\Index\Indexer->bulkInsertDocuments()              Indexer.php
Loupe\Loupe\Loupe->addDocument()                                       Loupe.php
```

### Steps to reproduce

1. Run on a PHP whose PDO_SQLite is < 3.35 (e.g. 3.34.1).
2. Add any document to an index (`$loupe->addDocument([...])`).
3. → syntax error near `RETURNING`.

### Suggested fix

Gate the `RETURNING` path on the runtime SQLite version and fall back when it is unavailable — i.e. implement the fallback the existing comment already anticipates. Detecting the version is cheap:

```php
$sqliteVersion = $connection->executeQuery('SELECT sqlite_version()')->fetchOne();
$supportsReturning = version_compare($sqliteVersion, '3.35.0', '>=');
```

When unsupported, run the upsert without `RETURNING` and obtain the needed columns via a follow-up `SELECT` on the unique key (or skip the returned-columns optimization for that path).

Alternatively (or in addition), documenting a hard minimum of **SQLite 3.35** and enforcing it at bootstrap (fail fast with a clear message) would at least surface the incompatibility at setup time instead of at the first index write.

### Notes

Happy to help test a patch against the affected host (IONOS, SQLite 3.34.1). Thanks for Loupe! 🙏

sigtrygg-space:
Fixes #325

First of all: thank you for Loupe! It's a fantastic library, and the fact that it runs anywhere PHP and SQLite run is exactly why this PR exists. 🙂

## Re: "SQLite >= 3.35.0 is enforced through composer.json"

I read your reply on #325, and I fully understand the constraint. I'd still like to propose this PR, because `composer.json` can only enforce the requirement on the machine where `composer install` runs, not on the machine where Loupe actually executes. On shared hosting these aren't necessarily the same: `vendor/` gets built locally or in CI (with a modern SQLite) and then deployed, or ships bundled with a CMS plugin. In the reported case, Loupe 0.13.21 was installed and running on the host despite the constraint and then crashed on the first index write.

So this PR deliberately does **not** touch `composer.json`: SQLite >= 3.35.0 stays the official, documented, fast-path requirement. It only adds a runtime fallback so that Loupe degrades gracefully instead of throwing on every write when the runtime SQLite is older, filling in the placeholder that was already in the code (`// Depending on the version, we could introduce fallback solutions here`). A runtime version check with a clear exception message would fail more loudly than the current syntax error, but it wouldn't help anyone actually deploying to such hosts — with the fallback, Loupe just works there.

## Problem

Since 0.13, `BulkUpserter::execute()` unconditionally uses the `RETURNING` clause, which requires SQLite >= 3.35.0 (2021-03-12). On hosts whose PHP is linked against an older SQLite, not uncommon on shared hosting, where the runtime environment often differs from the machine that built `vendor/`, every index write throws:

```
PDOException: SQLSTATE[HY000]: General error: 1 near "RETURNING": syntax error
```

Reported reproduction: IONOS shared hosting, PHP 8.4.22, PDO_SQLite 3.34.1 (one release below 3.35.0). Loupe 0.12 worked on the same host.

## Strategy

- `BulkUpserterFactory` now determines the runtime SQLite version once per engine (`SELECT sqlite_version()`, lazily cached) and passes it to each `BulkUpserter`.
- On SQLite >= 3.35.0 nothing changes: the modern `RETURNING` path stays the default, and the SQL it produces is identical to before — no performance regression for the common case.
- On older SQLite, the upsert is executed without `RETURNING`, and when returning columns are requested, the result is reconstructed with a follow-up `SELECT <returningColumns> FROM <table> WHERE (<uniqueColumns>) IN (VALUES ...)` over the unique-column values of the just-upserted rows. The result has the exact shape and keying the `Indexer` expects.

### Trade-off: change detection

The modern path's `RETURNING` omits rows skipped by the change-detecting column (`WHERE _hash IS NOT excluded._hash`). The fallback cannot cheaply detect which rows were skipped, so it conservatively returns **all** upserted rows. The write-skipping `WHERE` itself still applies on the fallback path, and the practical impact is small: `Indexer::addDocuments()` already pre-filters unchanged documents against the stored hashes before the bulk upsert, and re-processing is idempotent (`ConflictMode::Ignore` everywhere downstream). This is documented in the `executeLegacy()` docblock.

### Effective version floor

The fallback's follow-up `SELECT` uses row-value `IN (VALUES ...)` (SQLite >= 3.15.0), which is subsumed by the `INSERT ... ON CONFLICT` upsert syntax both paths already require (SQLite >= 3.24.0). So the fallback effectively supports SQLite 3.24.0–3.34.x.

## Tests

- Unit tests compare the modern and fallback paths on the same data for all three shapes the `Indexer` uses (update with change detection + returning, conflict-ignore with returning and differently-ordered unique columns across multiple chunks, no returning columns).
- An engine-level test drives indexing, re-indexing (unchanged + updated document), search, and deletion end-to-end with the fallback forced.
- Since CI's SQLite is >= 3.35, truly exercising the `< 3.35` branch needs the version check to be overridable: `BulkUpserterFactory` takes an optional `$sqliteVersion` constructor argument (auto-detected when `null`), so the fallback branch is testable regardless of the installed SQLite. The engine-level test swaps the factory via reflection to avoid widening `Engine`'s constructor.

`composer phpstan`, `composer cs-fixer` (ecs) and the unit/functional/slow suites are green.

## Deliberately out of scope

- As said above, `composer.json` and the README are untouched, 3.35.0 remains the official requirement.
- Phrase search separately emits `MATERIALIZED`/`NOT MATERIALIZED` CTE hints, which are also 3.35.0+. Fully supporting older SQLite end-to-end would need those gated as well — happy to do that as a follow-up if you want it, but I kept this PR minimal and localized to the `BulkUpserter`.

Thanks for taking a look!

### DIFF
--- a/src/Internal/Index/BulkUpserter/BulkUpserter.php
+++ b/src/Internal/Index/BulkUpserter/BulkUpserter.php
@@ -11,10 +11,17 @@ use Loupe\Loupe\Internal\Util;
 
 class BulkUpserter
 {
+    /**
+     * The RETURNING clause was only introduced in SQLite 3.35.0. For older versions, a fallback
+     * without RETURNING is used.
+     */
+    public const MIN_SQLITE_VERSION_FOR_RETURNING = '3.35.0';
+
     public function __construct(
         private Connection $connection,
         private BulkUpsertConfig $bulkUpsertConfig,
         private int $variableLimit,
+        private string $sqliteVersion,
     ) {
 
     }
@@ -62,12 +69,16 @@ class BulkUpserter
         $updateColumns = array_values(array_diff($this->bulkUpsertConfig->getRowColumns(), $this->bulkUpsertConfig->getUniqueColumns()));
         $chunkSize = max((int) round($this->variableLimit / \count($this->bulkUpsertConfig->getRowColumns()), 0, PHP_ROUND_HALF_DOWN), 1);
         $results = [];
+        $supportsReturning = version_compare($this->sqliteVersion, self::MIN_SQLITE_VERSION_FOR_RETURNING, '>=');
 
         foreach (Util::arrayChunk($this->bulkUpsertConfig->getRows(), $chunkSize) as $chunk) {
-            // Modern path: INSERT .. ON CONFLICT .. DO UPDATE [RETURNING]
-            $results = [...$results, ...$this->executeModern($chunk, $updateColumns)];
+            $chunkResults = $supportsReturning
+                // Modern path: INSERT .. ON CONFLICT .. DO UPDATE [RETURNING]
+                ? $this->executeModern($chunk, $updateColumns)
+                // Legacy path for SQLite < 3.35.0 without RETURNING support
+                : $this->executeLegacy($chunk, $updateColumns);
 
-            // Depending on the version, we could introduce fallback solutions here
+            $results = [...$results, ...$chunkResults];
         }
 
         return $results;
@@ -75,20 +86,20 @@ class BulkUpserter
 
     /**
      * @param array<array<int, mixed>> $rows
+     * @param array<int|string> $columnKeys
      * @param array<int<0, max>|string, mixed> $parameters
      */
-    private function buildValuesClause(array $rows, array &$parameters): string
+    private function buildTuples(array $rows, array $columnKeys, array &$parameters): string
     {
-        $columnKeys = array_keys($this->bulkUpsertConfig->getRowColumns());
-        $columnsCount = \count($columnKeys);
-
+        $tuple = $this->placeholdersRow(\count($columnKeys));
         $tuples = [];
+
         foreach ($rows as $row) {
             foreach ($columnKeys as $columnKey) {
                 $parameters[] = $row[$columnKey] ?? null;
             }
 
-            $tuples[] = $this->placeholdersRow($columnsCount);
+            $tuples[] = $tuple;
         }
 
         return implode(',', $tuples);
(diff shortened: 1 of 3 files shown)


Toflar:
Thanks for the PR, but I’m going to decline this.

I don’t plan to support SQLite versions older than 3.35.0. That minimum version was a deliberate choice and adding a legacy fallback path would create extra maintenance and testing burden that I’m not willing to take on.

Even if AI can help generate the implementation, I would still need to review, understand, test and maintain it long-term. And I simply don’t want that complexity in this project.

SQLite 3.35.0 is already more than five years old, so I consider it a reasonable minimum. If you need support for older SQLite versions, you’re welcome to maintain a fork for that use case.

sigtrygg-space:
Thanks for taking the time to review, and for the clear decision. It's a fair call.

Keeping the maintenance and testing surface small is exactly what makes a lean library like Loupe sustainable, and SQLite 3.35.0 being over five years old is a reasonable line to draw. In fact, I'm quite unhappy that the hoster does Not provide a newer version.

For anyone landing here from #325 with the same situation (runtime SQLite older than the machine that built `vendor/`, typically shared hosting), the options are:

- **Loupe 0.12.x** — worked in the reported setup and remains an option if you can live without the 0.13 improvements.
- **This PR's patch** stays available via https://github.com/loupe-php/loupe/pull/326.diff and can be applied locally, e.g. with `cweagans/composer-patches`.
- **The clean fix**: get your host onto a newer SQLite version - on some shared hosters, switching to a newer PHP version also brings a newer bundled `pdo_sqlite`, so that may be worth checking first.

Thanks again for Loupe!

Toflar:
Thanks for understanding! Shared hosting providers usually have some limitations which is fine. I suggest using [SEAL](https://php-cmsig.github.io/search/) for which I worked on a Loupe adapter too. We used that in Contao so basically our backend search is abstracted which means if you absolutely need the search feature and your provider does not fulfill the requirements for Loupe, you can just switch to any other supported adapter of SEAL and e.g. go with Meilisearch or Elasticsearch etc. with just a simple configuration adjustment.

Also regarding staying on Loupe v0.12: Just know that v0.12 has issues with concurrent writes (which SQLite does not support) and is just all in all way, way slower than 0.13 😊

sigtrygg-space:
Great suggestion, I'll have a look into SEAL, Thank you!

PR #337, opened 2026-07-21
Reduce database size using `WITHOUT ROWID`

ausi:
SQLite supports using clustered indexes using the `WITHOUT ROWID` keywords. This is especially helpful for tables with a composite primary key. See <https://sqlite.org/withoutrowid.html>

This shrinks the bench database from 216 MB by 50% down to 108 MB and reduces query time up to 26%.

```
+-------------+---------------------------+-----+------+-----+----------------+------------------+----------------+
| benchmark   | subject                   | set | revs | its | mem_peak       | mode             | rstdev         |
+-------------+---------------------------+-----+------+-----+----------------+------------------+----------------+
| SearchBench | benchExactQueryWithFacets |     | 3    | 5   | 9.750mb 0.00%  | 4.61ms -2.31%    | ±2.54% -18.46% |
| SearchBench | benchMultiWordQueries     |     | 1    | 5   | 10.079mb 0.00% | 101.17ms -14.20% | ±0.77% -72.79% |
| SearchBench | benchNegatedTermQueries   |     | 1    | 5   | 9.750mb 0.00%  | 11.33ms -5.75%   | ±3.69% +12.43% |
| SearchBench | benchPhraseGroupQueries   |     | 1    | 5   | 4.689mb 0.00%  | 21.62ms -1.60%   | ±1.73% -54.54% |
| SearchBench | benchPlainQuery           |     | 3    | 5   | 4.318mb 0.00%  | 5.09ms -7.20%    | ±3.43% +58.99% |
| SearchBench | benchSingleWordQueries    |     | 1    | 5   | 9.750mb 0.00%  | 22.34ms -26.01%  | ±4.23% -10.22% |
| SearchBench | benchTypoQueryWithFacets  |     | 3    | 5   | 9.750mb 0.00%  | 4.43ms -0.07%    | ±2.20% -32.42% |
| FilterBench | benchFilteredSortedSearch |     | 3    | 5   | 9.762mb 0.00%  | 15.17ms -2.40%   | ±3.28% -12.43% |
+-------------+---------------------------+-----+------+-----+----------------+------------------+----------------+
```

Sadly doctrine does not seem to support the `WITHOUT ROWID` table option for SQLite, but I could try to open a pull request for that there.

### DIFF
--- a/src/Internal/Index/IndexInfo.php
+++ b/src/Internal/Index/IndexInfo.php
@@ -641,7 +641,6 @@ class IndexInfo
         ;
 
         $table->setPrimaryKey(['term', 'document', 'attribute', 'position']);
-        $table->addIndex(['document']);
     }
 
     private function addTermsToSchema(Schema $schema): void
@@ -730,6 +729,28 @@ class IndexInfo
         // Schema changes invalidate query planner statistics; drop them before introspecting
         $connection->executeStatement('DROP TABLE IF EXISTS sqlite_stat1');
         $connection->executeStatement('DROP TABLE IF EXISTS sqlite_stat4');
+
+        $schemaDiff = $comparator->compareSchemas($schemaManager->introspectSchema(), $this->getSchema());
+
+        foreach ($schemaDiff->getCreatedTables() as $table) {
+            if (
+                in_array(
+                    $table->getObjectName()->getUnqualifiedName()->getValue(),
+                    [
+                        self::TABLE_NAME_MULTI_ATTRIBUTES_DOCUMENTS,
+                        self::TABLE_NAME_TERMS_DOCUMENTS,
+                        self::TABLE_NAME_PREFIXES_TERMS,
+                    ],
+                )
+            ) {
+                foreach ($connection->getDatabasePlatform()->getCreateTableSQL($table) as $statement) {
+                    if (str_starts_with($statement, 'CREATE TABLE')) {
+                        $statement .= ' WITHOUT ROWID';
+                    }
+                    $connection->executeStatement($statement);
+                }
+            }
+        }
 
         $schemaDiff = $comparator->compareSchemas($schemaManager->introspectSchema(), $this->getSchema());
         $schemaManager->alterSchema($schemaDiff);


Toflar:
Nice! I didn't know about that (I guess because other database engines do this by default but I see the BC issue mentioned in the SQLite docs 😊 ).

I can definitely confirm that this reduces the DB size by **50%** (🤯 ) and I can also confirm the search performance improvements between **15 and 26%** which is massive - so I defnitely want to get this working.
There are blockers though:

- We still support DBAL 3 so this has to be considered
- Removing the`document` index heavily regresses updates and deletes. Document replacement and deletion query this column directly. On my benchmark database, deleting one document’s term relations measured:
    - Existing index: 0.49 ms
    - `WITHOUT ROWID`, index removed: 39.7 ms 😟 
- Another issue is that as long as we do not contribute that to Doctrine, existing search indexes will not benefit from any of this optimizations. Doctrine’s schema comparison does not distinguish an existing rowid table from its `WITHOUT ROWID` equivalent.
- We must probably also  increase the `Engine::VERSION` constant to trigger a re-index on update.

ausi:
> * Removing the`document` index heavily regresses updates and deletes.

The `document` index alone takes around 65 MB of space. But I think we can improve that. I will look into it.

> * Doctrine’s schema comparison does not distinguish an existing rowid table from its `WITHOUT ROWID` equivalent.

I don’t know if this will ever be the case, even if `WITHOUT ROWID` support is added. There seems to be no way to change a regular table to a `WITHOUT ROWID` table, we have to drop the table and recreate it, and I don’t think doctrine supports such cases in general.

ausi:
> > * Removing the`document` index heavily regresses updates and deletes.
> 
> The `document` index alone takes around 65 MB of space. But I think we can improve that. I will look into it.

We could create another table like `CREATE TABLE documents_terms (document INTEGER NOT NULL, term INTEGER NOT NULL, PRIMARY KEY (document, term)) WITHOUT ROWID;` that acts as an index of documents to terms mapping and update the `SELECT` part from `WHERE document = ?` to `WHERE document = ? and term IN (SELECT term FROM documents_terms WERE document = ?)`. This should restore the performance for updates and deletes while only costing around 25 MB of storage. But it would be another table we have to maintain.

Should I try to implement that approach?

Toflar:
First of all, let's start by making the database size measurable: https://github.com/loupe-php/loupe/pull/338

Toflar:
While researching your findings, I found one very simple improvement on the relational table already that we can use in our current schema without any big changes and this already saves up quite some indexing space: https://github.com/loupe-php/loupe/pull/339

ausi:
> Sadly doctrine does not seem to support the `WITHOUT ROWID` table option for SQLite

There is a pull request now: https://github.com/doctrine/dbal/pull/7474

PR #341, opened 2026-07-23
Add covering index for term document searches

Toflar:
Compared with the `main` branch:

| Benchmark | Before | After | Change |
|---|---:|---:|---:|
| Multi-word queries | 111.68 ms | 102.84 ms | -7.9% |
| Phrase group queries | 21.43 ms | 20.30 ms | -5.3% |
| Plain query | 5.78 ms | 5.30 ms | -8.3% |
| Single-word queries | 34.29 ms | 25.92 ms | -24.4% |

All other benchmarks remain effectively unchanged.

But this would make the database even bigger... @ausi

### DIFF
--- a/src/Internal/Index/IndexInfo.php
+++ b/src/Internal/Index/IndexInfo.php
@@ -18,6 +18,8 @@ use Loupe\Loupe\Internal\Util;
 
 class IndexInfo
 {
+    public const INDEX_NAME_TERMS_DOCUMENTS_SEARCH = 'terms_documents_search';
+
     public const TABLE_NAME_DOCUMENTS = 'documents';
 
     public const TABLE_NAME_INDEX_INFO = 'info';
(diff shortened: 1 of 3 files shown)


src/Internal/Index/IndexInfo.php
```diff
@@ -642,6 +644,10 @@ private function addTermsToDocumentsRelationToSchema(Schema $schema): void
 
         $table->setPrimaryKey(['term', 'document', 'position']);
         $table->addIndex(['document']);
+        $table->addIndex(
+            ['term', 'document', 'attribute', 'position', 'folded'],
```
ausi, line 648:
```suggestion
            ['term', 'attribute', 'document', 'position', 'folded'],
```
Shouldn’t this be the other way around? If I search for term X in attribute Y the document is not known so in order to use the index for the attribute it has to appear before the document.

Toflar: Your ordering is better for the specific lookup `(term = X, attribute = Y)` because SQLite can narrow by both before scanning documents. But the order in this PR favors the other dominant paths: unrestricted searches group by `(term, document)` and phrase continuation - as they already know the document.
But feel free to bench 😊

Toflar:
The new covering index accounts for 77.89 MiB of additional storage but I think the performance improvement in searching justifies this change. After all, Loupe is all about searching. The size on the disk should be a factor but not the most important one. Once we can make use of `without_rowid`, we'll shrink down the physical size on disk again 😊

PR #342, opened 2026-07-23
Store attribute names in a separate table

ausi:
This would bring the `terms_documents` table down by 25% and should work great with the other pull requests as it also reduces the size of any index on `terms_documents.attribute`.

If you agree with this change I would try to update the query side as well. _(Currently only the index side is implemented)_


Toflar:
Sure, this will make the indexes smaller for sure :)

PR #345, opened 2026-07-30
Synonyms

daun:
(🥂 Cheers to 1.0! 🥂)

This adds a new feature: match documents by synonyms / aliases of the used query terms.

### Usage

Configure via `$configuration->withSynonyms()`.

 Each mapping is one-directional: the key is the query term, its values are the document terms it should also match.

 ```php
   $configuration = \Loupe\Loupe\Configuration::create()
       ->withSynonyms([
           'jacket' => ['parka', 'windbreaker'],
       ])
   ;
 ```

 Two-way synonyms need to be declared explicitly:

 ```php
   $configuration = \Loupe\Loupe\Configuration::create()
       ->withSynonyms([
           'couch' => ['sofa'],
           'sofa'  => ['couch'],

       ])
   ;
 ```

### Implementation

Synonyms are resolved at index time. During tokenization, a matching document token has the query terms added as additional token variants. The same mechanism already used for stems and prefixes.

There's no query-side cost, search latency is unaffected. The cost is a slightly larger index.

A document containing the queried term outranks one that only matches through a synonym.

### Limitations

Changing synonyms requires a reindex. The synonym map is part of the configuration hash, so `Loupe::needsReindex()` triggers after a change.

No resolution chains: `a → b` and `b → c` does not mean `a → c`. That relation must be declared explicitly.

Multi-word synonyms are not supported at this point since they throw up some more complicated questions.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -79,6 +79,36 @@ $configuration = \Loupe\Loupe\Configuration::create()
 ;
 ```
 
+## Synonyms
+
+Synonyms let you define terms that should additionally match other terms during search. They are read **query-side**
+and are **one-way**. In the example below, searching for `jacket` also finds documents containing `parka` or `windbreaker`,
+but searching for `parka` does NOT find generic jackets.
+
+```php
+$configuration = \Loupe\Loupe\Configuration::create()
+    ->withSynonyms([
+        'jacket' => ['parka', 'windbreaker'],
+    ])
+;
+```
+
+Two-way synonyms are expressed by listing both directions explicitly.
+
+```php
+$configuration = \Loupe\Loupe\Configuration::create()
+    ->withSynonyms([
+        'couch' => ['sofa'],
+        'sofa' => ['couch'],
+    ])
+;
+```
+
+A few things to keep in mind:
+
+- **Single-word only.** Every synonym source and target must be a single word. Multi-word entries are unsupported.
+- **Changing synonyms requires a reindex.** The synonyms are resolved during indexing and only take effect after a reindex.
+- **Literal matches rank above synonym matches**. A document containing the original term outranks one that only matches via a synonym.
 
 ## Minimum length for prefix search
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -81,9 +81,9 @@ $configuration = \Loupe\Loupe\Configuration::create()
 
 ## Synonyms
 
-Synonyms let you define terms that should additionally match other terms during search. They are read **query-side**
-and are **one-way**. In the example below, searching for `jacket` also finds documents containing `parka` or `windbreaker`,
-but searching for `parka` does NOT find generic jackets.
+Synonyms let terms match other terms during search. Each mapping is **one-way**: the key is a query term and its values
+are matching document terms. In this example, searching for `jacket` also finds documents containing `parka` or
+`windbreaker`, but searching for `parka` does not find generic jackets.
 
 ```php
 $configuration = \Loupe\Loupe\Configuration::create()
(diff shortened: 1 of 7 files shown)


daun:
@Toflar I think there's some issue with the code quality tools auto-fetching its binaries during installation. Maybe some race condition in a runner cache? 

```
Updating code quality tools …
[phpstan (composer)] 
[phpstan (composer)] In Filesystem.php line 261:
[phpstan (composer)]                                                                                
[phpstan (composer)]   /home/runner/.cache/composer/files/phpstan/phpstan does not exist and could  
[phpstan (composer)]    not be created:                                                             
[phpstan (composer)]                                                                                
[phpstan (composer)] 
[phpstan (composer)] update [--with WITH] [--prefer-source] [--prefer-dist] [--prefer-install PREFER-INSTALL] [--dry-run] [--dev] [--no-dev] [--lock] [--no-install] [--no-audit] [--audit-format AUDIT-FORMAT] [--no-security-blocking] [--no-blocking] [--no-autoloader] [--no-suggest] [--no-progress] [-w|--with-dependencies] [-W|--with-all-dependencies] [-v|vv|vvv|--verbose] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--strict-psr-autoloader] [--apcu-autoloader] [--apcu-autoloader-prefix APCU-AUTOLOADER-PREFIX] [--ignore-platform-req IGNORE-PLATFORM-REQ] [--ignore-platform-reqs] [--prefer-stable] [--prefer-lowest] [-m|--minimal-changes] [--patch-only] [-i|--interactive] [--root-reqs] [--bump-after-update [BUMP-AFTER-UPDATE]] [--] [<packages>...]
[phpstan (composer)] 
Error: Processes failed: phpstan (composer)
  ```

Toflar (commented):
I just re-ran the job and it worked 🤷‍♂️ 

I know you've implemented it like stop words but I'd argue that one probably has a lot more synonyms than stopwords, don't you think?

Maybe we should have an interface and ship an in-memory provider but I think fetching those lazily from some database might be a use case. Another use case might be storing them in the loupe DB itself. Wdyt?

daun:
> I know you've implemented it like stop words but I'd argue that one probably has a lot more synonyms than stopwords, don't you think?

Depends, but yeah. I didn't imagine hundreds of terms from a dictionary. More like a project-based thing, where editors alias specific search terms known to trip up users. Maybe a dozen terms, in my mind.

That said, I can see people pushing it, adding enough aliases to approach semantic search or something. We implemented something like this in a previous project, and the terms people entered were wild. It was mostly names of sponsors to make them show up in all searches, if I remember correctly. 🤠

> Maybe we should have an interface and ship an in-memory provider but I think fetching those lazily from some database might be a use case. Another use case might be storing them in the loupe DB itself.

I can see this moving out of the raw array syntax, sure. It ties a bit into something we've discussed before: persisting the configuration externally, so you don't need to pass it in every time you instantiate the engine. I vaguely recall a PR about storing config in a file? Like an index manager. In any case, if we do it for synonyms, we should do it for stop words too, with one consistent mechanism.

Toflar:
> In any case, if we do it for synonyms, we should do it for stop words too, with one consistent mechanism.

Yeah I think for both cases we can accept an array of strings for simplicity and an interface for more advanced settings. 
Both need to return their configuration hash identifier that we use to determine if re-indexing is necessary.

daun:
@Toflar Are you leaning more `array|WordCollectionInterface` or just `WordCollectionInterface` and a built-in `InMemoryWordCollection`? I find the latter one to be a bit cleaner, but let me know what you prefer and I can draft something.

Toflar:
The latter is definitely cleaner but I would probably not make it a collection anyway but a lazy lookup. If you have thousands of synonyms and stop words, you might want to load only what's needed🤔

PR #346, opened 2026-08-07
Force reindex when languages change

daun:
Changing the languages should probably trigger a reindex since it influences stemming and decompounding.


daun:
I'm not familiar with what the slow tests really assert, but there's been a GitHub outage last night, so the failing test might just be an aftermath of that...

> Test Failed (Loupe\Loupe\Tests\Slow\ConcurrencyTest::testManyConcurrentProcesses)
> Failed asserting that 499 is identical to 500.

Toflar (approved):

Toflar:
Thanks!

PR #367, opened 2026-09-06
Speed up primary-key iteration

daun:
This adds a special-case fast path to the browse API: primary-key-only iteration.

I've been trying to snapshot the existing document ids before and after an insert for proper cleanup of stale documents. Doing that I've noticed the browse API json-decodes every document, regardless of retrieved attributes, resulting in comparably slow ~~and memory-intensive~~ requests.

Public API is unchanged, it just internally branches on `withAttributesToRetrieve(['id'])` vs. all other requests.

**Benchmarks**

Added a benchmark for the browse API.

140 ms → 20 ms (-86%)
6MB → 4MB (-33%)

Correction: a prior benchmark overstated the memory reductions.

**Alternative**

As an alternative to an id-only fast path, one could consider using  `json_extract` to select only the columns worth fetching. That's more generic, but would require another benchmark and implementation.

### DIFF
--- a/src/Internal/Engine.php
+++ b/src/Internal/Engine.php
@@ -126,6 +126,14 @@ class Engine
         $this->maybeWrapStateSetIndexWithCache();
 
         try {
+            if (
+                '' === $parameters->getQuery()
+                && '' === $parameters->getFilter()
+                && [$this->configuration->getPrimaryKey()] === $parameters->getAttributesToRetrieve()
+            ) {
+                return $this->getConnection()->transactional(fn () => $this->browsePrimaryKeys($parameters));
+            }
+
             return (new Searcher($this, $this->filterParser, $parameters))->fetchResult();
         } catch (Exception $exception) {
             // If we need a re-index (e.g. schema has changed via an update from an old to a newer Loupe version)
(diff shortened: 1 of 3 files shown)


Toflar:
I think `json_extract` should be benched, don't you think? Could also affect regular search.
For the common sync case we could also provide some specialized `patch()` endpoint or so?

daun:
> I think `json_extract` should be benched, don't you think?

Right, good idea. I'll add a benchmark.

> For the common sync case we could also provide some specialized `patch()` endpoint or so?

What would the patch endpoint do? Or when would I use it? Not sure how it relates to the id iteration case :)

Toflar:
> What would the patch endpoint do?

Ah, I somehow interpreted this as being a solution to try and sync a Loupe index with another system. To update/patch what's not in the index yet. Should have said it differently: Instead of only optimizing the id retrieval, maybe we can also introduce a specific helper that would cover your use case even better :) 
But let's try `json_extract` first :)

daun:
Right, I'm just trying to get a list of all ids in the index to diff before and after an update. But I'm also very open to how exactly to get that :)

Toflar:
Let's try `json_extract` (and a special path when only the ID is requested) which should bring a nice perf benefit for all use cases :)

daun:
@Toflar I've done some tests locally and I'm seeing around a 20% speedup when using `json_extract`, but it comes at the cost of:

a) losing the original document key order (they would now be sorted alphabetically), and 
b) losing null values (so both missing keys and keys of null values are no longer part of the document)

I think those aren't really acceptable, especially in combination. What's your opinion, before I open a PR? 

The technical fix (restoring original key order and restoring null values after the fact) slows down small documents by quite a lot (50%), so that's not acceptable either.

I'm currently trying a few other things around indexes for faster browse pagination, will report back.

daun:
Just a note: the prior benchmark for this PR overstated the memory reduction quite a bit. It's more like 33% instead of 95%. Corrected and noted above.

daun:
I've added [a PR for a browse benchmark](https://github.com/loupe-php/loupe/pull/369) so we have exact numbers before continuing :)

daun:
Closing in favor of #370

PR #368, opened 2026-09-06
Respect sqlite max variable number for batch deletes

daun:
The document deletion endpoint is currently unbounded, i.e. passing in an array of 50K ids would crash for being above sqlite's max variable number of 32766. This change batches deletes to also use a lower number of variables. I've found the magic number of 5000 floating around, so I've formalized it into a private const `MAX_IDS_PER_QUERY` and re-used it there.


daun:
@Toflar I think the failed test is a GitHub actions outage, could you rerun it please?

Toflar (approved):

Toflar:
Thanks! :)

PR #370, opened 2026-09-07
Improve performance of document browsing API

daun:
## Changes

- Create Browser class for optimized browsing / document scanning
- Read documents table directly instead of going through the Searcher if neither query nor filter are used
- Skip reading the document body if only the primary key is requested
- Stream document pages instead of single-column fetch for memory reduction
- Note: on the fast path, the primary key's type is reconstructed from the schema

### Speed

  | Subject | Corpus | Before | After | Δ |
  |---|---|---:|---:|---:|
  | `benchBrowsePrimaryKey` | 31,944 × 510 B | 127.40 ms | 18.10 ms | **−85.79%** |
  | `benchBrowseWholeDocument` | 31,944 × 510 B | 127.46 ms | 90.16 ms | **−29.26%** |
  | `benchBrowseSubset` | 31,944 × 510 B | 127.46 ms | 92.82 ms | **−27.18%** |
  | `benchBrowseLargeSubset` | 20,000 × 4 KB | 561.70 ms | 533.15 ms | −5.08% |
  | `benchBrowseLargeWholeDocument` | 20,000 × 4 KB | 573.28 ms | 554.50 ms | −3.28% |
  | `benchBrowseSubsetFiltered` | 31,944 × 510 B | 222.03 ms | 215.87 ms | −2.77% |

  ### Memory

  | Subject | Before | After | Δ |
  |---|---:|---:|---:|
  | `benchBrowseSubset` | 2.207 MB | 1.603 MB | −27.37% |
  | `benchBrowsePrimaryKey` | 2.088 MB | 1.522 MB | −27.11% |
  | `benchBrowseLargeSubset` | 1.956 MB | 1.571 MB | −19.68% |
  | `benchBrowseWholeDocument` | 3.073 MB | 2.514 MB | −18.19% |
  | `benchBrowseLargeWholeDocument` | 6.187 MB | 5.806 MB | −6.16% |
  | `benchBrowseSubsetFiltered` | 2.560 MB | 2.560 MB | ±0.00% |

### DIFF
--- a/src/Internal/Search/Browser.php
+++ b/src/Internal/Search/Browser.php
@@ -1,0 +1,126 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Loupe\Loupe\Internal\Search;
+
+use Doctrine\DBAL\Query\QueryBuilder;
+use Loupe\Loupe\BrowseParameters;
+use Loupe\Loupe\BrowseResult;
+use Loupe\Loupe\Internal\Engine;
+use Loupe\Loupe\Internal\Filter\Parser;
+use Loupe\Loupe\Internal\Index\IndexInfo;
+use Loupe\Loupe\Internal\LoupeTypes;
+use Loupe\Loupe\Internal\Util;
+use Loupe\Loupe\SearchParameters;
+
+class Browser
+{
+    public function __construct(
+        private readonly Engine $engine,
+        private readonly Parser $filterParser,
+        private readonly BrowseParameters $parameters,
+    ) {
+    }
+
+    public function fetchResult(): BrowseResult
+    {
+        if (!$this->canScanDocuments()) {
+            return (new Searcher($this->engine, $this->filterParser, $this->parameters))->fetchResult();
+        }
+
+        return $this->engine->getConnection()->transactional(fn (): BrowseResult => $this->scanDocuments());
+    }
+
+    private function canScanDocuments(): bool
+    {
+        return '' === $this->parameters->getQuery()
+            && '' === $this->parameters->getFilter()
+            && $this->parameters->getOffset() >= 0
+            && $this->parameters->getLimit() >= 0;
+    }
+
+    /**
+     * @param array<string> $attributesToRetrieve
+     *
+     * @return array<array<string, mixed>>
+     */
+    private function fetchHits(QueryBuilder $queryBuilder, string $documentsAlias, array $attributesToRetrieve): array
+    {
+        $primaryKey = $this->engine->getConfiguration()->getPrimaryKey();
+
+        if ([$primaryKey] === $attributesToRetrieve) {
+            $ids = $queryBuilder
+                ->select($this->primaryKeyExpression($documentsAlias, $primaryKey))
+                ->setParameter('__loupe_pk_path', '$.'.$primaryKey)
+                ->fetchFirstColumn()
+            ;
+
+            return array_map(static fn (mixed $id): array => [$primaryKey => $id], $ids);
+        }
+
+        $documents = $queryBuilder->select($documentsAlias.'._document')->executeQuery()->iterateColumn();
+        $hits = [];
+
+        if (\in_array('*', $attributesToRetrieve, true)) {
+            foreach ($documents as $document) {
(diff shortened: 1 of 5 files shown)


src/Internal/Search/Browser.php
```diff
@@ -0,0 +1,126 @@
...
+use Loupe\Loupe\Internal\Filter\Parser;
+use Loupe\Loupe\Internal\Index\IndexInfo;
+use Loupe\Loupe\Internal\LoupeTypes;
+use Loupe\Loupe\Internal\Util;
+use Loupe\Loupe\SearchParameters;
+
+class Browser
+{
+    public function __construct(
+        private readonly Engine $engine,
```
Toflar, line 20: Why not just decorating `Searcher` right away?

src/Internal/Search/Browser.php
```diff
@@ -0,0 +1,126 @@
...
+
+        return $hits;
+    }
+
+    private function primaryKeyExpression(string $documentsAlias, string $primaryKey): string
+    {
+        return match ($this->engine->getIndexInfo()->getDocumentSchema()[$primaryKey] ?? null) {
+            LoupeTypes::TYPE_STRING => $documentsAlias.'._user_id',
+            LoupeTypes::TYPE_NUMBER => \sprintf('CAST(%s._user_id AS NUMERIC)', $documentsAlias),
+            default => \sprintf('json_extract(%s._document, :__loupe_pk_path)', $documentsAlias),
```
Toflar, line 87: That cannot happen, can it? And even if it can, we would need to check for support of `json_extract` because afaik we cannot assume they are always there because prior to 3.38 you could disable JSON support.

src/Internal/Search/Browser.php
```diff
@@ -0,0 +1,126 @@
...
+        private readonly Engine $engine,
+        private readonly Parser $filterParser,
+        private readonly BrowseParameters $parameters,
+    ) {
+    }
+
+    public function fetchResult(): BrowseResult
+    {
+        if (!$this->canScanDocuments()) {
+            return (new Searcher($this->engine, $this->filterParser, $this->parameters))->fetchResult();
```
Toflar, line 29: I don't like that we now have a `Browser` that then does delegate to `Searcher` that then does not search but browse. This is confusing.

Toflar (changes requested):

src/Internal/Search/Browser.php
daun: I had it in `Engine` previously, as a special case. That felt strange because it was generating all kinds of SQL. When you say decorate, you mean what exactly — a trait for browsing tacked onto `Searcher`?

src/Internal/Search/Browser.php
daun: How do you mean, that cannot happen? That a document id might come back as string or number?

src/Internal/Search/Browser.php
daun: Would you prefer it all in the `Searcher` class? As noted above, a special trait for the non-search browsing would certainly make sense to lift that confusion.

Toflar: Yeah I'm not against splitting the browsing and searching process completely but then it has to be done all the way and not delegate a part of it to the searcher again because that I find confusing.

src/Internal/Search/Browser.php
Toflar: I don't see why we need this case. The primary key is always stored in `_user_id` - I'm probably confused about what this part of the code is supposed to do.

daun: This part is about returning the primary key in its original type. The primary key in `_user_id` is coerced to a string before storage (`$userId = (string) $document[$primaryKey]`), so we don't know the original type just from returning that. If we're fine with returning a string even though document ids were passed in as numbers, we can scratch this part. But it didn't feel quite right to me.

src/Internal/Search/Browser.php
daun: Understand, so it's either inline it again or rip it out all the way. I'll give it a go, see which makes most sense.

PR #371, opened 2026-09-07
Allow configuring JSON backend

daun:
Add configuration options for defining the json backend to use for encoding/decoding. Users can supply a callable for encoding (indexing) and/or decoding (searching). Useful in environments where you're able install PHP extensions of faster json-decode implementations.

```php
    ->withJsonEncoder(static fn ($array) => fastjson_encode($array))
    ->withJsonDecoder(static fn ($json) => fastjson_decode($json, true));
```

### Benchmarks

Seeing around 11% speedup for search and 4% for indexing when using [fastjson](https://github.com/iliaal/fastjson).  
No change with default configuration, of course.

#### Search / decoding

  | Backend           | Time | Delta | Peak mem |
  |-------------------|---------------:|----------:|----------:|
  | native `ext-json` |        15.09 ms |        —  |  ~60.3 MB |
  | `fastjson` |        13.45 ms |   -10.9%  |  ~59.0 MB |

#### Indexing / encoding

  | Backend            | Time        | Delta | Peak mem |
  |--------------------|------------:|------:|---------:|
  | native `ext-json`  | 55,137.45 ms|   —   | 101.2 MB |
  | `fastjson` | 52,841.08 ms| -4.2% | 102.0 MB |

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -40,6 +40,8 @@ final class Configuration
 
     private \Closure|null $jsonDecoder = null;
 
+    private \Closure|null $jsonEncoder = null;
+
     private int $maxQueryTokens = 10;
 
     private int $maxTotalHits = 1000;
@@ -257,6 +259,11 @@ final class Configuration
         return $this->jsonDecoder;
     }
 
+    public function getJsonEncoder(): \Closure|null
+    {
+        return $this->jsonEncoder;
+    }
+
     public function getLogger(): LoggerInterface|null
     {
         return $this->logger;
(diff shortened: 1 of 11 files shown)


Toflar (commented):
So this is not just allowing the backend but actually picking another one if available, right? :) I'm not against that, because if you have it installed, sure, why not.
But I don't really like the `\Closure` interface. I know it's simple but probably we should just have a `JsonInterface` with `encode`, `decode` and `validate` (while we're on int).
Would probably give us proper error handling as well while for a `\Closure` this is undefined now.

Also, using `fastjson` will use almost 1.8 times the memory of native PHP `json_decode()` - yes it is faster but it comes at a cost apparently.

daun:
Yes, it allows picking another one, it doesn't do any checks on its own. Validate is a great idea!

I also assumed `fastjson` to be higher memory usage, but cold-indexing the whole 32K movie set comes in at 102MB as opposed to 101MB for native ext-json. Maybe our test documents just aren't complex enough to trigger the memory-hungry part of it? Might also explain why the speed gains are rather modest in the end. So the general recommendation might be to only reach for this if a) you have complex documents and b) you have lots of memory :)

daun:
Rereading you comment again, just to clarify: this PR does not check for alternative backends. It doesn't know about fastjson, simdjson, etc. It requires a conscious decision and configuration by the consumer, not least because of implications for memory consumption.

